PR #5, opened 2026-02-18
perf(waymo-converter): optimize lidar processing (23x speedup)

zgojcic:
Addresses severe performance bottlenecks in the Waymo-to-NCore V4 lidar conversion pipeline, reducing per-frame processing from ~7.3s to ~0.3s (verified bit-exact output).

Key optimizations:

1. Replace O(N*S) broadcast ray matching with O(N+S) index lookup The primary/secondary return matching used an N×S boolean matrix via broadcasting (~4.4s/frame for ~130K×40K points). Replaced with a direct index-map array for sub-millisecond matching.

2. Fast-path protobuf parser for MatrixFloat/MatrixInt32 Python protobuf ParseFromString is extremely slow for large packed repeated fields (~1M+ floats). Added a targeted wire-format parser that locates the packed data field and extracts it via np.frombuffer, bypassing Python repeated-field construction (~2.2s → ~0.07s/frame).

3. Cache top-pose TF tensor across first/second return calls convert_range_image_to_cartesian was called twice per frame and rebuilt the [H,W,4,4] top-pose tensor each time. Now cached via a dict passed between calls.

4. Batch per-label rolling-shutter projection Label centroids are now batched into a single call to world_points_to_sensor_angles_shutter_pose instead of individual per-label calls (~146ms → ~71ms/frame).

5. Replace TF reshape with numpy for single 4x4 pose matrix frame.pose.transform was being reshaped via tf.reshape/tf.constant instead of a simple numpy reshape.

### DIFF
--- a/tools/data_converter/waymo/utils.py
+++ b/tools/data_converter/waymo/utils.py
@@ -16,6 +16,9 @@
 #
 # Modifications Copyright 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
+import zlib
+
+from dataclasses import dataclass
 from typing import Optional
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


janickm:
CI is failing due to formatting - `bazel run format` should help

janickm:
@zgojcic sorry I messed around a little - I rebased this branch

janickm:
[applied formatting in the latest commit]

janickm:
[let me know once I can do a full review]

zgojcic:
Also switched to a batch processing now. Should be good to review. There is a small differnece in terms of directions and angles (not sure where exactly it comes from) but its max value on my clip is 10e-4. 

Maybe you can try and compare on the clip you were running @ph03

janickm:
[rebased on `main`]

janickm:
Verified that results are consistent also on the dataset I'm using `10724020115992582208_7660_400_7680_400` - just have a few minor updates I'll still push

tools/data_converter/waymo/utils.py
```diff
@@ -25,6 +28,129 @@
...
+    while pos < end:
+        val, pos = _decode_varint(buf, pos)
+        val = val & 0xFFFFFFFF
+        if val > 0x7FFFFFFF:
+            val -= 0x100000000
+        values.append(val)
+    return values
+
+
+def _parse_matrix_proto(raw_compressed: bytes, dtype: type) -> _ParsedMatrix:
```
janickm, line 74: magic :)

janickm (approved):
verified locally that results are consistent relative to previous changes (also verified lidar-model-based projections)

janickm:
@zgojcic the commit is not GPG signed and can't be merged as is (resulting in "Commits must have verified signatures."), which is required for the [DCO](https://github.com/NVIDIA/ncore?tab=contributing-ov-file#developer-certificate-of-origin) we need for this repo

We have instructions on how to set this up once for your account

https://github.com/NVIDIA/ncore?tab=contributing-ov-file#setting-up-gpg-signing

Do you want to check if this works for you? You'll have to create the key, register it locally in git (once) and once in your GH account, and then `commit --amend` (with `-S` if you don't enable this by default) the commit + force push

janickm:
[I'm still updating based on offline notes]

janickm:
removed unused `_decode_zigzag` but kept current `_decode_variant` as it's 2x faster compared to `google.protobuf.internal.decoder._DecodeVarint`

up to you @zgojcic to sign the commit

janickm (approved):

PR #42, opened 2026-03-09
feat(ncore_vis): Add up direction and metadata colors to ncore_vis.

michael-shelley:
## Description

Adds up direction to dropdown and allows arbitrary metadata to color lidar points. This will improve colmap support.

## Type of Change

- [ ] Documentation update

## Checklist

- [ ] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [ ] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [ ] My commits are GPG-signed
- [ ] I have added tests that prove my fix is effective or my feature works
- [ ] Code is formatted (`bazel run //:format`)
- [ ] I have updated documentation as needed
- [ ] My changes include SPDX license headers on all new files

### DIFF
--- a/tools/ncore_vis/components/lidar.py
+++ b/tools/ncore_vis/components/lidar.py
@@ -76,13 +76,10 @@ class LidarComponent(VisualizationComponent):
     def __init__(self, client, data_loader, renderer):
         super().__init__(client, data_loader, renderer)
 
-        # Per-sensor generic_data fields that can drive point color.
-        self._metadata_color_fields: Dict[str, List[str]] = {}
-
-        # scan lidar data for metadata that can be used to colorize points
-        for lidar_id in self.data_loader.lidar_ids:
-            metadata_fields = self._scan_metadata_color_fields(lidar_id)
-            self._metadata_color_fields[lidar_id] = metadata_fields
+        # Scan for per-sensor generic_data fields that can drive point color.
+        self._metadata_color_fields: Dict[str, List[str]] = {
+            lidar_id: self._scan_metadata_color_fields(lidar_id) for lidar_id in self.data_loader.lidar_ids
+        }
 
     def create_gui(self, tab_group: viser.GuiTabGroupHandle) -> None:  # noqa: C901
         self._enabled: bool = True
@@ -235,18 +232,24 @@ class LidarComponent(VisualizationComponent):
         A field qualifies if it has exactly 1, 3, or 4 values per point.
         """
         sensor = self.data_loader.get_lidar_sensor(lidar_id)
-        ts = sensor.frames_timestamps_us[0, 1].item()
+
+        if sensor.frames_count == 0:
+            return []
+
         n_points = sensor.get_frame_ray_bundle_count(0)
+
         if n_points == 0:
             return []
 
         color_metadata_fields: List[str] = []
-        for name in sensor.get_frame_generic_data_names(ts):
-            data = np.asarray(sensor.get_frame_generic_data(ts, name))
+
+        for name in sensor.get_frame_generic_data_names(0):
+            data = np.asarray(sensor.get_frame_generic_data(0, name))
             if data.ndim == 1 and data.shape[0] == n_points:
                 color_metadata_fields.append(name)
             elif data.ndim == 2 and data.shape[0] == n_points and data.shape[1] in [1, 3, 4]:
                 color_metadata_fields.append(name)
+
         return color_metadata_fields
 
     def _bind_lidar_callbacks(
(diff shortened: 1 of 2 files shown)


tools/ncore_vis/components/lidar.py
```diff
@@ -216,6 +222,26 @@ def on_reference_frame_change(self, interval_us: HalfClosedInterval) -> None:
     # Internal helpers
     # ------------------------------------------------------------------
 
+    def _scan_metadata_color_fields(self, lidar_id: str) -> List[str]:
+        """Return generic_data field names from frame 0 whose shape qualifies for coloring.
+
+        A field qualifies if it has exactly 1, 3, or 4 values per point.
+        """
+        qualifying: List[str] = []
```
janickm: exit early (return empty list) for the case that there is no frame 0

tools/ncore_vis/components/lidar.py
```diff
@@ -535,3 +564,38 @@ def _color_model_element(self, lidar_id: str, frame: int, n_points: int, column:
...
         return (rgba[:, :3] * 255.0).astype(np.uint8)
+
+    def _color_metadata_field(self, lidar_id: str, frame: int, field_name: str, n_points: int) -> np.ndarray:
+        """Color by a generic_data metadata field.
+
+        Single-channel fields are treated as intensity (grayscale).
+        Three-channel fields are treated as RGB.
+        Four-channel fields are treated as RGBA (alpha-premultiplied onto black).
+
+        Float data is assumed to be in [0, 1] and scaled to uint8.
```
janickm: we can't assume this, e.g., if we have a "speed of point" (in m/s), then that would not be normalized) - I guess we need to have a "normalized" yes/no selector for a single scalar field

tools/ncore_vis/components/lidar.py
```diff
@@ -535,3 +564,38 @@ def _color_model_element(self, lidar_id: str, frame: int, n_points: int, column:
...
+        Integer data is scaled linearly by its dtype maximum.
+        """
+        try:
+            sensor = self.data_loader.get_lidar_sensor(lidar_id)
```
janickm, line 593: instantiating a sensor is costly - let's try to re-use these globally (likely requires a rework of this class?)

tools/ncore_vis/components/lidar.py
```diff
@@ -535,3 +564,38 @@ def _color_model_element(self, lidar_id: str, frame: int, n_points: int, column:
...
+            ts = sensor.frames_timestamps_us[frame, 1].item()
+            data = np.asarray(sensor.get_frame_generic_data(ts, field_name))
+        except Exception:
```
janickm: why could this fail? catching generic exceptions like this can hide issues, or?

janickm (changes requested):
thanks @michael-shelley ! just left a few minor comments

janickm:
[rebased on `main`]

tools/ncore_vis/components/lidar.py
michael-shelley: I've normalized by default for non-uint8s. Do you think it's sufficient to normalize over just a single frame?

tools/ncore_vis/components/lidar.py
michael-shelley: Moved to init

tools/ncore_vis/components/lidar.py
michael-shelley: Changed to KeyError - missing data at the given timestamp could raise this error, right?

tools/ncore_vis/components/lidar.py
janickm: thx

tools/ncore_vis/components/lidar.py
janickm: added this now - resolving

tools/ncore_vis/components/lidar.py
janickm: I think it's a key error on missing timestamp, but it's an index error when accessing the `frames_timestamps_us` array - I pushed an update that exits early on missing frames, should be good now

janickm (approved):

tools/ncore_vis/components/lidar.py
janickm: I guess that sounds good (we can still adapt later if needed, let's not over-complicate :))

janickm:
Actually there is a bug when running on AV data now..

  File "/home/janickm/.cache/bazel/_bazel_janickm/9d22350bcd24cc2d81624fb73f1b4d67/execroot/_main/bazel-out/k8-fastbuild/bin/tools/ncore_vis/ncore_vis.runfiles/_main/ncore/impl/data/v4/compat.py", line 232, in get_frame_generic_data_names
    self.frames_timestamps_us[frame_index, FrameTimepoint.END.value].item()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: index 126081 is out of bounds for axis 0 with size 199

checking

janickm:
Pushed a fix, seems to be working now 

image

janickm:
merging

PR #49, opened 2026-03-16
colmap: Don't include downsampled images by default

janickm:
This pull request updates the default behavior for handling downsampled images in the COLMAP data converter. The primary change is that downsampled images are no longer included as additional cameras by default, to avoid downstream issues in 3DGRUT/GSplat. This affects both the codebase and the documentation to ensure consistent behavior and clear communication to users.

**Default behavior changes:**

* Changed the default value of the `include_downsampled_images` configuration option in the `ColmapConverter4Config` class from `True` to `False`, so downsampled images are not included as additional cameras unless explicitly specified. (`tools/data_converter/colmap/converter.py`)
* Updated the default value of the `--include-downsampled-images` CLI option to `False` to match the new default behavior. (`tools/data_converter/colmap/converter.py`)

**Documentation updates:**

* Updated the documentation to reflect that the default value for `--include-downsampled-images` is now `False` instead of `True`. (`docs/conversions/colmap/colmap.rst`)

### DIFF
--- a/docs/conversions/colmap/colmap.rst
+++ b/docs/conversions/colmap/colmap.rst
@@ -136,7 +136,7 @@ subdirectory is treated as a separate sequence.
      - Prefix prepended to COLMAP integer camera IDs to form NCore sensor IDs
        (e.g. ``camera1``)
    * - ``--include-downsampled-images BOOL``
-     - ``True``
+     - ``False``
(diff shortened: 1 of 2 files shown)


michael-shelley (approved):
LGTM

PR #74, opened 2026-04-16
perf: IndexedTarStore: cache tail read to avoid duplicate I/O for in-range keys

eattia-nvidia:
Keep the byte buffer from the index tail read and serve __getitem__ from it when a tar payload lies entirely in that range, so overlapping keys near the end of the archive do not trigger another read on remote backends.

Add a regression test that wraps the file object and asserts no further read() calls when the tail spans the whole file.

## Type of Change

- [ ] Documentation update

## Checklist

- [ ] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [ ] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [ ] My commits are GPG-signed
- [ ] I have added tests that prove my fix is effective or my feature works
- [ ] Code is formatted (`bazel run //:format`)
- [ ] I have updated documentation as needed
- [ ] My changes include SPDX license headers on all new files

### DIFF
--- a/ncore/impl/data/stores.py
+++ b/ncore/impl/data/stores.py
@@ -60,6 +60,11 @@ class IndexedTarStore(Store):
         Default 1 MiB size fits typical use cases, but may be increased if the compressed
         index size is expected to be larger or decreased if remote access chunk sizes are smaller.
 
+    Read mode retains the byte range loaded for the index (the tail read) as an in-memory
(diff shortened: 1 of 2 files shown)


janickm (changes requested):
thanks @eattia-nvidia - I amended the commit with a few refinements, merging

PR #104, opened 2026-05-08
fix(bazel): patch protobuf for NixOS PATH compatibility

janickm:
## Summary

- Patches protobuf@29.1 `embed_edition_defaults` rule to add `use_default_shell_env = True` to its `ctx.actions.run_shell()` call
- Without this, the action runs with an empty `PATH`, causing `sed`/`cp` not found errors on NixOS (where tools live in `/nix/store`, not `/usr/bin`)

## Root Cause

Protobuf's `_embed_edition_defaults_impl` in `editions/defaults.bzl` uses `ctx.actions.run_shell()` without `use_default_shell_env = True`. In Bazel 8, this creates a hermetic action environment with no `PATH` at all — `--action_env=PATH` and `--noincompatible_strict_action_env` only apply to actions that opt in via `use_default_shell_env`. On NixOS there are no tools at standard FHS paths like `/usr/bin/sed`, so the build fails.

## Changes

| File | Description |
|------|-------------|
| `MODULE.bazel` | `single_version_override` for protobuf with the patch |
| `bazel/protobuf/BUILD.bazel` | Exports the patch file |
| `bazel/protobuf/protobuf_use_default_shell_env.patch` | One-line patch adding `use_default_shell_env = True` |

## Testing

- `bazel build //...` — passes (912 actions)
- `bazel test --config=no-gpu //...` — 25/28 pass (3 failures are pre-existing Python 3.8 + PyTorch `libtorch_cpu.so` incompatibility, unrelated)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -33,6 +33,18 @@ bazel_dep(name = "bazel_skylib", version = "1.7.1")
 bazel_dep(name = "rules_proto", version = "7.1.0")
 bazel_dep(name = "platforms", version = "1.0.0")
 
+## protobuf
+# Patch protobuf to add use_default_shell_env=True to embed_edition_defaults.
+# Without this, ctx.actions.run_shell runs with an empty PATH, which breaks on
+# NixOS where tools like sed/cp live in /nix/store and are not at /usr/bin.
+single_version_override(
+    module_name = "protobuf",
+    patch_strip = 1,
+    patches = [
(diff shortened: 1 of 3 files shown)


janickm:
Closing in favor of #105 which takes the cleaner approach: instead of patching protobuf 29.1, it bumps to protobuf 30.0 (which fixes the NixOS issue upstream) and also removes the archived `rules_proto` dependency entirely.

Verified: full build, all 28 tests, and Waymo dataset conversion all pass with the new approach.

Issue #90, opened 2026-05-05
ROS2 Support


hasghaier:
### Problem Statement

How to convert from ROS2 bag file to ncore data ?

### Proposed Solution

Hello, 
Does any one already had the idea or the initiative to create an ncore converter from ROS2 bag files ?
Is there any suggestions on where to start from ?
Is it planned to extend the converters to other type of datasets, like Kitti for example ?

Thanks in advance. Great repo by the way 👍

janickm:
Thanks for the question! Here's our assessment:

## ROS2 bag support:

A general-purpose ROS2 bag converter isn't feasible — rosbag2 is a container format (MCAP/SQLite3 + CDR serialization), not a dataset specification. The bag is self-describing (embeds topic names and message types), but the actual sensor layout, topic naming, and data semantics are entirely user-defined. A converter would require dataset-specific knowledge of which topics map to which sensors, making a generic solution impractical. Users with ROS2 bags would need to write a custom converter tailored to their specific recording setup. This is better implemented using ncore as a library following the existing converters as examples.

## KITTI support:

Supporting KITTI raw data is viable and we're working towards it. One prerequisite is that NCore requires pre-computed ego poses (SE3 transforms), while KITTI provides raw GPS/IMU (OXTS) data that must be converted to poses via Mercator projection. We also want to preserve the raw OXTS measurements alongside the derived poses.

To support this, we're first adding component-level generic data storage to the V4 format (#103), which allows attaching auxiliary numpy arrays (like raw GPS/IMU time-series) to any component without defining a new component type. The KITTI converter will be built on top of that change.

For the OXTS-to-pose conversion itself, there's a well-established reference implementation in [pykitti](https://github.com/utiasSTARS/pykitti) that me might use as a dependency (license-wise it looks permissible, will need to check the details).

janickm:
Just added support to convert KITTI datasets to NCore-v4 with #109 - closing this for now

janickm:
Example:

<img width="2161" height="945" alt="Image" src="https://github.com/user-attachments/assets/9224f2bd-1b57-48e5-9e0d-fd31e0c98b06" />

PR #108, opened 2026-05-12
fix(ncore_vis): handle grayscale camera images

janickm:
## Summary

Use `get_frame_image()` (PIL) instead of `get_frame_image_array()` (numpy) in the camera visualizer, converting to RGB before passing to viser. This fixes an `IndexError` on grayscale cameras (e.g., KITTI image_00/image_01) where viser expects a 3-channel array but gets a 2D `(H, W)` array.

## Changes

| File | Description |
|------|-------------|
| `tools/ncore_vis/components/camera.py` | Convert PIL image to RGB before `np.asarray()` |

Single-line logic change, no API modifications.

### DIFF
--- a/tools/ncore_vis/components/camera.py
+++ b/tools/ncore_vis/components/camera.py
@@ -787,8 +787,11 @@ class CameraComponent(VisualizationComponent):
             )
             self._poses[camera_id] = pose_handle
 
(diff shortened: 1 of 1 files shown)


janickm:
Related to adding kitty support (#90)

PR #116, opened 2026-05-15
feat(deps): make torch an optional dependency, only required for sensor evaluation

janickm:
## Summary

- Make `torch` an optional dependency -- users who only need data reading (`ncore.data`) no longer need PyTorch installed
- `ncore.sensors` now uses lazy imports via `__getattr__`, importing torch only when a sensor class is accessed
- Clear `ImportError` message guides users to install `ncore[sensors]` if torch is missing
- Wheel updated: torch moved from `requires` to `extra_requires = {"sensors": ["torch"]}`
- All data tests pass without torch at import time; all sensor tests pass with torch available

### DIFF
--- a/ncore/BUILD.bazel
+++ b/ncore/BUILD.bazel
@@ -62,10 +62,14 @@ py_wheel(
         "zarr>=2.12.0,<3.0.0",
         "cbor2",
         "scipy",
-        "torch",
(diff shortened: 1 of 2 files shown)


janickm:
Postponed -- needs more design work around packaging boundaries before making torch optional.

PR #117, opened 2026-05-15
chore: add from __future__ import annotations to all Python files

janickm:
## Summary

- Add `from __future__ import annotations` to 31 Python files that were missing it
- Covers `ncore/data/`, `ncore/sensors/`, `ncore/impl/`, `tools/`, and test files
- Aligns all files with the project coding convention specified in AGENTS.md


janickm:
Unnecessary -- only needed where forward references or newer syntax require it, not globally.

PR #119, opened 2026-05-15
refactor(sensors): extract shared rolling-shutter iteration logic into common utility

janickm:
## Summary

- Extract nearly-identical rolling-shutter iterative solver logic from `camera.py` and `lidar.py` into shared utility in `common.py`
- New `rolling_shutter_iterate()` function handles pose interpolation, convergence checking, and iteration loop
- Camera and lidar pass sensor-specific callbacks (projection, time computation, convergence metric, state update)
- Pure refactor -- no behavioral changes, all existing sensor tests pass on both Python 3.8 and 3.11

### DIFF
--- a/ncore/impl/sensors/camera.py
+++ b/ncore/impl/sensors/camera.py
@@ -28,6 +28,7 @@ from ncore.impl.common.util import map_optional, unpack_optional
 from ncore.impl.data import types
 from ncore.impl.sensors.common import (
     BaseModel,
+    RollingShutterSolver,
     eval_poly_horner,
     eval_poly_inverse_horner_newton,
     rotmat_to_unitquat,
@@ -224,6 +225,24 @@ class BivariateWindshieldModel(ExternalDistortionModel):
             self.order_theta,
             self.poly_eval_2d,
         )
+
+
+class _CameraRollingShutterProjector(RollingShutterSolver.Projector):
+    """Camera-specific rolling-shutter projector.
+
+    Convergence is measured by mean absolute change in relative frame time (derived from
+    projected image coordinates) between iterations.
+    """
+
+    def __init__(self, camera_model: CameraModel) -> None:
+        self._model = camera_model
+
+    def project(self, sensor_points: torch.Tensor) -> RollingShutterSolver.ProjectionResult:
+        result = self._model.camera_rays_to_image_points(sensor_points)
+        return RollingShutterSolver.ProjectionResult(projected=result.image_points, valid_flag=result.valid_flag)
+
+    def relative_frame_time(self, projected: torch.Tensor) -> torch.Tensor:
+        return self._model.image_points_relative_frame_times(projected)
 
 
 class CameraModel(BaseModel, ABC):
@@ -434,8 +453,7 @@ class CameraModel(BaseModel, ABC):
         start_timestamp_us: Optional[int] = None,
         end_timestamp_us: Optional[int] = None,
         max_iterations: int = 10,
-        stop_mean_error_px: float = 1e-3,
-        stop_delta_mean_error_px: float = 1e-5,
+        stop_t_delta: float = 1e-4,  # ~0.1 px for 1080p
         return_T_world_sensors: bool = False,
         return_valid_indices: bool = False,
         return_timestamps: bool = False,
@@ -457,8 +475,7 @@ class CameraModel(BaseModel, ABC):
             start_timestamp_us,
             end_timestamp_us,
             max_iterations,
-            stop_mean_error_px,
-            stop_delta_mean_error_px,
+            stop_t_delta,
             return_T_world_sensors,
             return_valid_indices,
             return_timestamps,
@@ -554,8 +571,7 @@ class CameraModel(BaseModel, ABC):
         start_timestamp_us: Optional[int] = None,
         end_timestamp_us: Optional[int] = None,
         max_iterations: int = 10,
-        stop_mean_error_px: float = 1e-3,
-        stop_delta_mean_error_px: float = 1e-5,
+        stop_t_delta: float = 1e-4,  # ~0.1 px for 1080p
         return_T_world_sensors: bool = False,
         return_valid_indices: bool = False,
         return_timestamps: bool = False,
@@ -589,15 +605,13 @@ class CameraModel(BaseModel, ABC):
             start_timestamp_us = int(start_timestamp_us)
             end_timestamp_us = int(end_timestamp_us)
 
-        # Always perform transformation using start pose
-        image_points_start = self.camera_rays_to_image_points(
-            (T_world_sensor_start[:3, :3] @ world_points.transpose(0, 1) + T_world_sensor_start[:3, 3, None]).transpose(
-                0, 1
-            )
-        )
-
-        # Global-shutter special case - no need for rolling-shutter compensation, use projections from start-pose as single available pose
+        # Global-shutter special case - no need for rolling-shutter compensation
         if self.shutter_type == types.ShutterType.GLOBAL:
+            image_points_start = self.camera_rays_to_image_points(
+                (
+                    T_world_sensor_start[:3, :3] @ world_points.transpose(0, 1) + T_world_sensor_start[:3, 3, None]
+                ).transpose(0, 1)
+            )
             return_var = self.WorldPointsToImagePointsReturn(
                 image_points=(
                     image_points_start.image_points[image_points_start.valid_flag]
@@ -618,125 +632,53 @@ class CameraModel(BaseModel, ABC):
                 )
             return return_var
 
-        # Do initial transformations using both start and mean pose to determine all candidate points and take union of valid projections as iteration starting points
-        image_points_end = self.camera_rays_to_image_points(
-            (T_world_sensor_end[:3, :3] @ world_points.transpose(0, 1) + T_world_sensor_end[:3, 3, None]).transpose(
-                0, 1
-            )
+        iter_result = RollingShutterSolver.solve(
+            world_points=world_points,
+            T_world_sensor_start=T_world_sensor_start,
+            T_world_sensor_end=T_world_sensor_end,
+            max_iterations=max_iterations,
+            stop_t_delta=stop_t_delta,
+            projector=_CameraRollingShutterProjector(self),
         )
 
-        valid = image_points_start.valid_flag | image_points_end.valid_flag  # union of valid image points
-        init_image_points = image_points_end.image_points
-        init_image_points[image_points_start.valid_flag] = image_points_start.image_points[
-            image_points_start.valid_flag
-        ]  # this prefers points at the start-of-frame pose over end-of-frame points
-        # - the optimization will determine the final timestamp for each point
-
-        # Exit early if no point projected to a valid image point
-        if not valid.any():
-            return_var = self.WorldPointsToImagePointsReturn(
-                image_points=(
-                    torch.empty((0, 2), dtype=self.dtype, device=self.device)
-                    if not return_all_projections
-                    else init_image_points
-                )
-            )
-            if return_T_world_sensors:
-                return_var.T_world_sensors = torch.empty((0, 4, 4), dtype=self.dtype, device=self.device)
-            if return_valid_indices:
-                return_var.valid_indices = torch.empty((0,), dtype=torch.int64, device=self.device)
-            if return_timestamps:
-                return_var.timestamps_us = torch.empty((0,), dtype=torch.int64, device=self.device)
-            return return_var
-
-        # Convert the start and end rotation matrix to quaternions for subsequent interpolations
-        world_sensor_s_quat = rotmat_to_unitquat(T_world_sensor_start[None, :3, :3])  # [1, 4]
-        world_sensor_e_quat = rotmat_to_unitquat(T_world_sensor_end[None, :3, :3])  # [1, 4]
-
-        # For valid image points, compute the new timestamp and project again
-        image_points_rs_prev = init_image_points[valid, :]
-        mean_error_px = 1e12
-
-        # Pre-compute values that are constant across iterations
-        n_valid = int(valid.sum().item())
-        s_quat_expanded = world_sensor_s_quat.expand(n_valid, -1)
-        e_quat_expanded = world_sensor_e_quat.expand(n_valid, -1)
-        trans_start = T_world_sensor_start[:3, 3]  # [3]
-        trans_end = T_world_sensor_end[:3, 3]  # [3]
-
-        for _ in range(max_iterations):
-            t = self.image_points_relative_frame_times(image_points_rs_prev)
-
-            rot_rs = unitquat_to_rotmat(unitquat_slerp(s_quat_expanded, e_quat_expanded, t))  # [n_valid, 3, 3]
-
-            # Use broadcasting for translation interpolation
-            trans_rs = (1 - t)[..., None] * trans_start + t[..., None] * trans_end
-
-            cam_rays_rs = (torch.bmm(rot_rs, world_points[valid, :, None]) + trans_rs[..., None]).squeeze(-1)
-            image_points_rs = self.camera_rays_to_image_points(cam_rays_rs)
-
-            # Compute mean error of projections that are still valid now and check if we are still
-            # making progress relative to previous iteration
-            if (
-                abs(
-                    mean_error_px
-                    - (
-                        mean_error_px := torch.linalg.norm(
-                            image_points_rs.image_points[image_points_rs.valid_flag]
-                            - image_points_rs_prev[image_points_rs.valid_flag],
-                            dim=1,
-                        ).mean()
-                    )
-                )
-                <= stop_delta_mean_error_px
-            ):
-                break
-
-            # Check if error bound was reached
-            if mean_error_px <= stop_mean_error_px:
-                break
-
-            image_points_rs_prev = image_points_rs.image_points
+        projection_valid = iter_result.projection_valid
+        projection_init = iter_result.projection_init
+        projection = iter_result.projection
+        rot_rs = iter_result.rot_rs
+        trans_rs = iter_result.trans_rs
+        t = iter_result.t
 
         # We always return image points
-        return_var = self.WorldPointsToImagePointsReturn(
-            image_points=image_points_rs.image_points[image_points_rs.valid_flag]
-        )
+        return_var = self.WorldPointsToImagePointsReturn(image_points=projection)
 
         if return_T_world_sensors:
-            # Generate the output matrix
-            trans_matrices = torch.empty(
-                (int(image_points_rs.valid_flag.sum().item()), 4, 4), dtype=self.dtype, device=self.device
-            )
-            trans_matrices[:, :3, 3] = trans_rs[image_points_rs.valid_flag]
-            trans_matrices[:, :3, :3] = rot_rs[image_points_rs.valid_flag, ...]
+            # Assemble [n, 4, 4] rigid transformation matrices from rotation and translation
+            n = projection.shape[0]
+            trans_matrices = torch.empty((n, 4, 4), dtype=self.dtype, device=self.device)
+            trans_matrices[:, :3, :3] = rot_rs
+            trans_matrices[:, :3, 3] = trans_rs
             trans_matrices[:, 3] = torch.tensor([0, 0, 0, 1], device=self.device, dtype=self.dtype)
-
             return_var.T_world_sensors = trans_matrices
 
         if return_valid_indices:
-            # Combine validity flags
-            # (valid_rs represents a strict logical subset of full valid flags, so no logical operation required)
-            valid[torch.argwhere(valid).squeeze()] = image_points_rs.valid_flag
-            return_var.valid_indices = torch.argwhere(valid).squeeze(1)
+            return_var.valid_indices = torch.argwhere(projection_valid).squeeze(1)
 
         if return_timestamps:
             return_var.timestamps_us = (
                 cast(int, start_timestamp_us)
-                + (
-                    t[image_points_rs.valid_flag, None] * (cast(int, end_timestamp_us) - cast(int, start_timestamp_us))
-                ).to(torch.int64)
+                + (t[:, None] * (cast(int, end_timestamp_us) - cast(int, start_timestamp_us))).to(torch.int64)
             ).squeeze(1)
 
         if return_all_projections:
-            if not return_valid_indices:
-                valid[torch.argwhere(valid).squeeze()] = image_points_rs.valid_flag
-                valid_indices = torch.argwhere(valid).squeeze(1)
-            else:
-                valid_indices = unpack_optional(return_var.valid_indices)
-
-            return_var.image_points = init_image_points
-            return_var.image_points[valid_indices] = image_points_rs.image_points[image_points_rs.valid_flag]
+            valid_indices = (
+                unpack_optional(return_var.valid_indices)
+                if return_valid_indices
+                else torch.argwhere(projection_valid).squeeze(1)
+            )
+            return_var.image_points = projection_init
+            return_var.image_points[valid_indices] = projection.to(
+                dtype=projection_init.dtype, device=projection_init.device
+            )
 
         return return_var
 
@@ -1070,9 +1012,9 @@ class CameraModel(BaseModel, ABC):
         t = self.image_points_relative_frame_times(image_points)
 
         # Use broadcasting instead of repeat() for translation interpolation
-        trans_start = T_sensor_world_start[:3, 3]  # [3]
-        trans_end = T_sensor_world_end[:3, 3]  # [3]
-        world_position_rs = (1 - t)[..., None] * trans_start + t[..., None] * trans_end  # [n_image_points, 3]
+        transl_start = T_sensor_world_start[:3, 3]  # [3]
+        transl_end = T_sensor_world_end[:3, 3]  # [3]
+        world_position_rs = (1 - t)[..., None] * transl_start + t[..., None] * transl_end  # [n_image_points, 3]
 
         R_sensor_world_rs = unitquat_to_rotmat(
             unitquat_slerp(
--- a/ncore/impl/sensors/common.py
+++ b/ncore/impl/sensors/common.py
@@ -13,7 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Union, cast
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Literal, Optional, Protocol, Union, cast
 
 import numpy as np
 import torch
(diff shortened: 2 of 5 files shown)


ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,231 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
     quat = quat / torch.norm(quat, dim=-1, keepdim=True)
 
     return quat
+
+
+# --- Rolling-shutter iteration utilities ---
+
+
+class RollingShutterSolver:
+    """Iterative rolling-shutter solver shared by camera and lidar projection.
```
janickm: Make this generic over general 2d shutter sensors

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,231 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+
+        projected: torch.Tensor
+        valid_flag: torch.Tensor
+
+    @dataclass
+    class Result:
+        """Final result of the rolling-shutter iterative solver.
+
+        Attributes:
+            rot_rs: Interpolated rotation matrices [n_valid, 3, 3]
```
janickm: Would it make sense to make these n, not n_valid to simplify the api / make use is the mask?

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,231 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+            ----------
+            projected : torch.Tensor
+                Projected values (image points or sensor angles) [n, 2]
+
+            Returns
+            -------
+            torch.Tensor
+                Relative frame times in [0, 1] range [n]
+            """
+            ...
```
janickm: could we call this / include "relative_frame_time" somehow?

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,231 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+            """Compute scalar convergence metric for the current iteration.
+
+            Parameters
+            ----------
+            projection : RollingShutterSolver.ProjectionResult
+                Current projection result
+            projected_prev : torch.Tensor
+                Previous iteration's projected values [n, 2]
+            t : torch.Tensor
+                Current relative frame times [n]
```
janickm: Variables are called inconsistent

ncore/impl/sensors/lidar.py
```diff
@@ -27,13 +27,48 @@
...
+        return (t[projection.valid_flag] - time_prev[projection.valid_flag]).abs().mean().item()
+
+    def update_state(
+        self,
+        prev_projected: torch.Tensor,
+        prev_time: torch.Tensor,
+        projection: RollingShutterSolver.ProjectionResult,
+        t: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        prev_projected[projection.valid_flag] = projection.projected[projection.valid_flag]
```
janickm: this feels inconsistent to the camera version - maybe this is the more general version that camera should also be using / we shouldn't even require such an update_state specialiazation?

ncore/impl/sensors/common.py
janickm: Done -- updated docstring to describe this as a "generic iterative solver for 2D shutter sensors", removed camera/lidar specifics.

ncore/impl/sensors/common.py
janickm: Fixed -- docstrings now document shapes as `[n, ...]` since the solver just processes whatever points it receives. The caller is responsible for the valid/invalid split before calling solve().

ncore/impl/sensors/common.py
janickm: Done -- renamed to `compute_relative_frame_time`, matching the existing sensor method names (`image_points_relative_frame_times`, `sensor_angles_relative_frame_times`).

ncore/impl/sensors/common.py
janickm: Fixed -- now consistently uses `t` for current and `t_prev` for previous throughout the solver, Protocol, and all implementations.

ncore/impl/sensors/lidar.py
janickm: Agreed -- removed `update_state` from the Protocol entirely. The solver now hardcodes the selective update (only valid projections update state) for both camera and lidar. This is the safer general approach: prevents runaway extrapolation when projections land outside the sensor FOV.

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,205 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+        ----------
+        world_points_valid : torch.Tensor
+            World points [n, 3]
+        s_quat_expanded : torch.Tensor
+            Start rotation quaternion expanded to [n, 4]
+        e_quat_expanded : torch.Tensor
+            End rotation quaternion expanded to [n, 4]
+        trans_start : torch.Tensor
+            Start translation [3]
+        trans_end : torch.Tensor
```
janickm: I don't like these names with s/e for `quat` and _start/_end for "trans" - also, why is translation not dependend on `n` - why are quaternions "expaned"? can you please clean this up?

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,205 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
     quat = quat / torch.norm(quat, dim=-1, keepdim=True)
 
     return quat
+
+
+# --- Rolling-shutter iteration utilities ---
+
+
+class RollingShutterSolver:
+    """Generic iterative solver for 2D shutter sensors.
```
janickm: There seems to be no strict reason to make this 2d only, or? could you generalize this solver to N-d shutter sensors (perhaps there is some use-case at some point) - just call this D or P as the dimension of the projection-space

ncore/impl/sensors/common.py
janickm: align this with the names in `Result`

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,205 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+            Start rotation quaternion expanded to [n, 4]
+        e_quat_expanded : torch.Tensor
+            End rotation quaternion expanded to [n, 4]
+        trans_start : torch.Tensor
+            Start translation [3]
+        trans_end : torch.Tensor
+            End translation [3]
+        initial_projected : torch.Tensor
+            Initial projected values [n, 2]
+        initial_relative_time : torch.Tensor
```
janickm: why do we use relative_time here an `t` elsewhere?

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,205 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+            Initial relative frame times [n]
+        max_iterations : int
+            Maximum number of iterations
+        stop_error : float
+            Absolute error threshold for early stopping
+        stop_delta_error : float
+            Delta error threshold (change between iterations) for early stopping
+        initial_error : float
```
janickm: move "initial" error up / before stop-error / stop-delta

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,205 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+            World points [n, 3]
+        s_quat_expanded : torch.Tensor
+            Start rotation quaternion expanded to [n, 4]
+        e_quat_expanded : torch.Tensor
+            End rotation quaternion expanded to [n, 4]
+        trans_start : torch.Tensor
+            Start translation [3]
+        trans_end : torch.Tensor
+            End translation [3]
+        initial_projected : torch.Tensor
```
janickm: could we make this part of the solver to do the initial projection?

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,205 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+            Protocol implementation providing sensor-specific projection behaviour
+
+        Returns
+        -------
+        RollingShutterSolver.Result
+            Final interpolated poses, times, and projection results
+        """
+        projected_prev = initial_projected
+        t_prev = initial_relative_time
+        mean_error = initial_error
```
janickm: you don't know that it's a "mean" error (this depends on the projector to choose)

ncore/impl/sensors/common.py
janickm: Fixed -- renamed to `quat_start`/`quat_end` for consistency with `trans_start`/`trans_end`. Translation stays `[3]` because the start/end poses are single rigid transforms applied to all n points (the per-point interpolation happens internally via slerp).

ncore/impl/sensors/common.py
janickm: Done -- docstrings now use `[n, D]` where D is the projection dimension. The code was already dimension-agnostic (pure tensor ops), so only documentation needed updating. Class docstring says "rolling-shutter sensors" without specifying 2D.

ncore/impl/sensors/common.py
janickm: Fixed -- renamed `initial_relative_time` to `t_init` for consistency with `t`/`t_prev` used in the loop.

ncore/impl/sensors/common.py
janickm: Done -- `initial_error` now comes before `stop_error`/`stop_delta_error` in the parameter list.

ncore/impl/sensors/lidar.py
janickm: Kept as a parameter (renamed to `projected_init`). The callers compute this via a union of start+end pose projections (warm-start heuristic), which is sensor-specific logic that does not belong in the generic solver. The solver just needs an initial guess to start iterating from.

ncore/impl/sensors/common.py
janickm: Also renamed `mean_error` to `error` -- the solver does not assume the metric is a mean, that is up to the projector.

ncore/impl/sensors/common.py
janickm: Fixed -- renamed to `quat_start`/`quat_end` for consistency with `trans_start`/`trans_end`. The quaternions are `[n, 4]` (caller expands them); translation is `[3]` because start/end are single rigid poses for all n points (per-point interpolation happens via slerp internally).

ncore/impl/sensors/common.py
janickm: Done -- docstrings now use `[n, D]` for the projection dimension. The code was already dimension-agnostic (pure tensor ops). Class docstring says "rolling-shutter sensors" without restricting to 2D.

ncore/impl/sensors/common.py
janickm: Fixed -- renamed `initial_relative_time` to `t_init` for consistency with `t`/`t_prev` used in the loop body.

ncore/impl/sensors/common.py
janickm: Done -- `initial_error` now comes before `stop_error`/`stop_delta_error` in the parameter list.

ncore/impl/sensors/common.py
janickm: Kept as a parameter (renamed to `projected_init`). Callers compute this via a union of start+end pose projections (warm-start heuristic that prefers start-of-frame values). This is sensor-specific initialization logic that does not belong in the generic solver -- the solver just needs an initial guess to start iterating from.

ncore/impl/sensors/common.py
janickm: Fixed -- renamed to just `error`. The solver makes no assumption about what the metric represents; that is entirely up to the projector's `convergence_error()` implementation.

ncore/impl/sensors/lidar.py
```diff
@@ -512,77 +536,53 @@ def world_points_to_sensor_angles_shutter_pose(
...
+        iter_result = RollingShutterSolver.solve(
+            world_points_valid=world_points[valid, :],
+            quat_start=s_quat_expanded,
+            quat_end=e_quat_expanded,
+            trans_start=trans_start,
+            trans_end=trans_end,
+            projected_init=initial_angles[valid, :],
+            t_init=relative_time[valid].clone(),
+            max_iterations=max_iterations,
+            initial_error=0.5,
```
janickm: move the original comment here on "# initialize the value to an expected value of random association [0,1]"

ncore/impl/sensors/lidar.py
```diff
@@ -512,77 +536,53 @@ def world_points_to_sensor_angles_shutter_pose(
...
-            if mean_relative_time_error <= stop_mean_relative_time_error:
-                break
+        iter_result = RollingShutterSolver.solve(
+            world_points_valid=world_points[valid, :],
+            quat_start=s_quat_expanded,
+            quat_end=e_quat_expanded,
+            trans_start=trans_start,
+            trans_end=trans_end,
+            projected_init=initial_angles[valid, :],
+            t_init=relative_time[valid].clone(),
```
janickm: is cloning really necessary?

ncore/impl/sensors/lidar.py
```diff
@@ -27,13 +27,37 @@
...
+    RollingShutterSolver,
     rotmat_to_unitquat,
     to_torch,
     unitquat_slerp,
     unitquat_to_rotmat,
 )
 
 
+class _LidarRollingShutterProjector:
+    """Lidar-specific rolling-shutter projector."""
```
janickm: indicate that the error is based on relative time changes / convergence

ncore/impl/sensors/camera.py
```diff
@@ -226,6 +227,36 @@ def undistort_camera_rays(self, camera_rays: torch.Tensor) -> torch.Tensor:
         )
 
 
+class _CameraRollingShutterProjector:
+    """Camera-specific rolling-shutter projector."""
```
janickm: indicate that the error is based on projection coordinate convergence

ncore/impl/sensors/lidar.py
janickm: Done -- added inline comment: `# initialize to expected value of random association in [0, 1]`

ncore/impl/sensors/lidar.py
janickm: Removed -- boolean indexing (`tensor[mask]`) already returns a copy in PyTorch, so `.clone()` was redundant.

ncore/impl/sensors/lidar.py
janickm: Done -- docstring now reads: "Convergence is measured by mean absolute change in relative frame time between iterations."

ncore/impl/sensors/camera.py
janickm: Done -- docstring now reads: "Convergence is measured by mean pixel displacement between iterations."

ncore/impl/sensors/camera.py
```diff
@@ -649,94 +657,40 @@ def world_points_to_image_points_shutter_pose(
...
+        return_var = self.WorldPointsToImagePointsReturn(image_points=projection)
 
         if return_T_world_sensors:
             # Generate the output matrix
-            trans_matrices = torch.empty(
-                (int(image_points_rs.valid_flag.sum().item()), 4, 4), dtype=self.dtype, device=self.device
-            )
-            trans_matrices[:, :3, 3] = trans_rs[image_points_rs.valid_flag]
-            trans_matrices[:, :3, :3] = rot_rs[image_points_rs.valid_flag, ...]
+            trans_matrices = torch.empty((projection.shape[0], 4, 4), dtype=self.dtype, device=self.device)
```
janickm: minor: could be simpler `torch.eye()` (saves one assignment below?) - or is there a simpler concat call?

ncore/impl/sensors/camera.py
```diff
@@ -618,22 +628,20 @@ def world_points_to_image_points_shutter_pose(
...
         )
 
-        valid = image_points_start.valid_flag | image_points_end.valid_flag  # union of valid image points
-        init_image_points = image_points_end.image_points
-        init_image_points[image_points_start.valid_flag] = image_points_start.image_points[
-            image_points_start.valid_flag
-        ]  # this prefers points at the start-of-frame pose over end-of-frame points
-        # - the optimization will determine the final timestamp for each point
+        projection_valid = iter_result.projection_valid
+        init_image_points = iter_result.projection_init
```
janickm: minor: rename init_image_points -> projection_init for readibility

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,211 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+        -------
+        RollingShutterSolver.Result
+            Final interpolated poses, times, projection results, validity mask,
+            and initial projections for all input points.
+        """
+        # Project all world points from start and end poses
+        sensor_rays_start = (
+            T_world_sensor_start[:3, :3] @ world_points.T + T_world_sensor_start[:3, 3, None]
+        ).T  # [n, 3]
+        sensor_rays_end = (T_world_sensor_end[:3, :3] @ world_points.T + T_world_sensor_end[:3, 3, None]).T  # [n, 3]
```
janickm: we should call these sensor_points_start/end and also update the names in the `projector.project()` method (only rename variables, the projectors usually can handle both points / rays)

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,211 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+        # Filter to valid points for iteration
+        world_points_valid = world_points[valid, :]
+        n_valid = world_points_valid.shape[0]
+
+        # Early return if no valid points
+        if n_valid == 0:
+            return RollingShutterSolver.Result(
+                rot_rs=torch.empty((0, 3, 3)),
+                trans_rs=torch.empty((0, 3)),
+                t=torch.empty(0),
```
janickm: is there a dtype we should be setting for the empty ones?

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,211 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+        )
+
+        for _ in range(max_iterations):
+            rot_rs = unitquat_to_rotmat(unitquat_slerp(quat_start, quat_end, t))  # [n_valid, 3, 3]
+
+            # Translation interpolation via broadcasting
+            trans_rs = (1 - t)[..., None] * transl_start + t[..., None] * transl_end
+
+            # Project sensor rays at interpolated pose
+            sensor_rays = (torch.bmm(rot_rs, world_points_valid[:, :, None]) + trans_rs[..., None]).squeeze(-1)
```
janickm: same, should be sensor_points (or sensor_points_rs)

ncore/impl/sensors/common.py
```diff
@@ -283,3 +286,211 @@ def unitquat_slerp(quat_s: torch.Tensor, quat_e: torch.Tensor, t: torch.Tensor,
...
+                t[projection_result.valid_flag] = t_new
+            else:
+                t_delta = 0.0
+
+            if t_delta < stop_t_delta:
+                break
+
+        # Combine initial validity (projected from start/end pose) with final iteration validity
+        # into a single [n] mask over all input points
+        projection_valid = valid.clone()
```
janickm, line 707: is cloning really required here?

ncore/impl/sensors/lidar.py
```diff
@@ -27,13 +27,28 @@
...
 )
 
 
+class _LidarRollingShutterProjector(RollingShutterSolver.Projector):
+    """Lidar-specific rolling-shutter projector."""
+
+    def __init__(self, lidar_model: RowOffsetStructuredSpinningLidarModel) -> None:
+        self._model = lidar_model
+
+    def project(self, sensor_rays: torch.Tensor) -> RollingShutterSolver.ProjectionResult:
```
janickm: see other comments - should be sensor_points

ncore/impl/sensors/lidar.py
```diff
@@ -470,130 +484,42 @@ def world_points_to_sensor_angles_shutter_pose(
...
+        return_var = self.WorldPointsToSensorAnglesReturn(sensor_angles=projection)
 
         if return_T_world_sensors:
             # Generate the output matrix
-            trans_matrices = torch.empty(
-                (int(sensor_angles.valid_flag.sum().item()), 4, 4), dtype=self.dtype, device=self.device
-            )
-            trans_matrices[:, :3, 3] = trans_rs[sensor_angles.valid_flag]
-            trans_matrices[:, :3, :3] = rot_rs[sensor_angles.valid_flag, ...]
+            trans_matrices = torch.empty((projection.shape[0], 4, 4), dtype=self.dtype, device=self.device)
```
janickm: same on torch.eye maybe

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        principal_point=np.array([959.5, 539.5], dtype=np.float32),
+        reference_poly=FThetaCameraModelParameters.PolynomialType.PIXELDIST_TO_ANGLE,
+        pixeldist_to_angle_poly=np.array([0.0, 1.0 / 500.0], dtype=np.float32),
+        angle_to_pixeldist_poly=np.array([0.0, 500.0], dtype=np.float32),
+        max_angle=float(np.radians(80)),
+    )
+    return FThetaCameraModel(params, device=device, dtype=dtype)
+
+
+def _rotation_matrix_y(angle_rad: float) -> np.ndarray:
```
janickm: you could use scipy for this

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+# =============================================================================
+# Section 1: Solver unit tests (mock projector)
+# =============================================================================
+
+
+class TestRollingShutterSolver(unittest.TestCase):
+    """Unit tests for RollingShutterSolver.solve using mock projectors."""
+
+    def setUp(self) -> None:
+        self.device = torch.device("cpu")
```
janickm: parameterize the test over both cpu and cuda

janickm: actually all tests that run torch

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        self.n_points = 10
+        # World points in front of sensor (z > 0)
+        torch.manual_seed(42)
+        self.world_points = torch.randn(self.n_points, 3, device=self.device, dtype=torch.float32)
+        self.world_points[:, 2] = self.world_points[:, 2].abs() + 1.0
+
+        # Identity poses
+        self.T_start = torch.eye(4, device=self.device, dtype=torch.float32)
+        self.T_end = torch.eye(4, device=self.device, dtype=torch.float32)
```
janickm: using identity poses is too simple for a rolling-shutter solver - this needs to be effective poses - is this intentional for this test case / different for the other ones? we should have both rotation and translation motion tested

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        self.dtype = torch.float32
+        self.camera = _make_ftheta_camera(self.device, self.dtype)
+        self.height = 1080
+        self.width = 1920
+
+    def test_self_consistency_fixed_point(self) -> None:
+        """Project world points with shutter compensation and verify fixed-point property.
+
+        For each valid output point (u, v) with time t*, the fundamental correctness
+        check is: t* == v / (height - 1) within tolerance.
```
janickm, line 332: mention the used shutter direction

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        world_points = torch.randn(50, 3, device=self.device, dtype=self.dtype)
+        world_points[:, 2] = world_points[:, 2].abs() + 5.0  # ensure in front of camera
+
+        T_start = _make_pose(np.eye(3, dtype=np.float32), np.array([0.0, 0.0, 0.0], dtype=np.float32))
+        T_end = _make_pose(
+            _rotation_matrix_y(np.radians(5)),
+            np.array([0.1, 0.0, 0.0], dtype=np.float32),
+        )
+
+        result = self.camera.world_points_to_image_points_shutter_pose(
```
janickm, line 347: add a variant that also checks return_timestamps and / or return_all_projections

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+            world_points,
+            T_start,
+            T_end,
+            max_iterations=20,
+            stop_t_delta=1e-7,
+            return_valid_indices=True,
+        )
+
+        image_points = result.image_points
+        valid_indices = result.valid_indices
```
janickm, line 360: why is this not used / checked?

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        T_identity = np.eye(4, dtype=np.float32)
+
+        # Rolling-shutter projection with no motion
+        rs_result = self.camera.world_points_to_image_points_shutter_pose(
+            world_points,
+            T_identity,
+            T_identity,
+            max_iterations=10,
+            stop_t_delta=1e-6,
+            return_valid_indices=True,
```
janickm, line 421: not used?

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        )
+
+        # Static projection: transform points to camera frame then project
+        cam_rays = world_points  # identity pose means cam_rays == world_points
+        static_result = self.camera.camera_rays_to_image_points(cam_rays)
+
+        # Compare valid points from both methods
+        # The rolling-shutter method returns only valid image points
+        # The static method returns all with a validity flag
+        static_valid_points = static_result.image_points[static_result.valid_flag]
```
janickm, line 428: make sure that there are some invalid points also and that this is consistent (could be behind camera or so)

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+
+        result = self.camera.world_points_to_image_points_shutter_pose(
+            world_point,
+            T_start,
+            T_end,
+            max_iterations=10,
+            stop_t_delta=1e-6,
+        )
+
+        # Single point on optical axis should be valid
```
janickm, line 461: please check validity then - add the return_valid_indices flag or so

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+            dtype=self.dtype,
+        )
+        T_identity = np.eye(4, dtype=np.float32)
+
+        result = self.camera.world_points_to_image_points_shutter_pose(
+            world_points,
+            T_identity,
+            T_identity,
+            max_iterations=10,
+            stop_t_delta=1e-6,
```
janickm, line 499: same - check all possible return values for expected / known values (return timestamps, valid flags etc.)

ncore/impl/sensors/rolling_shutter_test.py
janickm: also timestamps etc.

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def test_point_valid_at_start_invalid_at_end(self) -> None:
+        """A point inside FOV at start pose but outside at end pose is still processed."""
+        # Use a large rotation that moves the camera such that the point leaves FOV
+        # Point at moderate angle from optical axis
+        world_point = torch.tensor([[2.0, 0.0, 5.0]], device=self.device, dtype=self.dtype)
+
+        T_start = np.eye(4, dtype=np.float32)
+        # Large rotation around Y: 60 degrees -- the point will be outside FOV at end
+        T_end = _make_pose(_rotation_matrix_y(np.radians(60)), np.zeros(3, dtype=np.float32))
```
janickm: make sure this is the case by doing a local projection

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+
+        result = self.camera.world_points_to_image_points_shutter_pose(
+            world_point,
+            T_start,
+            T_end,
+            max_iterations=10,
+            stop_t_delta=1e-6,
+            return_valid_indices=True,
```
janickm, line 547: never used - check this output.

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+            np.array([0.5, 0.0, 0.0], dtype=np.float32),
+        )
+
+        result = self.camera.world_points_to_image_points_shutter_pose(
+            world_points,
+            T_start,
+            T_end,
+            max_iterations=20,
+            stop_t_delta=1e-6,
+            return_valid_indices=True,
```
janickm, line 636: never used? check it

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,490 @@
...
+        )
+
+        # Should have at least some valid projections
+        self.assertGreater(result.image_points.shape[0], 0)
+
+        # Self-consistency: for valid output points, t* == row / (height - 1)
+        image_points = result.image_points
+        rows = image_points[:, 1]
+        expected_t = rows / (self.height - 1)
+        actual_t = self.camera.image_points_relative_frame_times(image_points)
```
janickm, line 654: relative time can also be returned with return_timestamps - make use of this

ncore/impl/sensors/common.py
janickm: Done -- the solver is now fully dimension-agnostic (uses `[n, D]` throughout). Class docstring says 'rolling-shutter sensors' without restricting to 2D.

ncore/impl/sensors/common.py
janickm: Done -- the solver now operates on all `n` input points internally. `Result.projection_valid` is an `[n]` boolean mask; outputs (`rot_rs`, `trans_rs`, `t`, `projection`) are pre-filtered to `[n_final]` only truly-valid points.

ncore/impl/sensors/common.py
janickm: Done -- the method is now `compute_relative_frame_time` in the Projector protocol.

ncore/impl/sensors/common.py
janickm: Fixed -- all variables now use consistent naming throughout (`t` for time, `sensor_points` for transformed points, etc.).

ncore/impl/sensors/rolling_shutter_test.py
```diff
@@ -0,0 +1,567 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+import unittest
+
+import numpy as np
+import torch
+
+from scipy.spatial.transform import Rotation
```
janickm, line 28: when using scipy, you need to update BUILD.bazel

ncore/impl/sensors/lidar.py
janickm: Agreed -- removed `update_state` from the Protocol entirely. The solver now handles this internally (only valid projections update time). Both camera and lidar use the same generic path.

ncore/impl/sensors/common.py
janickm: Resolved -- the solver now accepts `T_world_sensor_start/end` as 4x4 poses and handles all decomposition (quaternion conversion, expansion) internally. Callers never see quaternions or translations.

ncore/impl/sensors/common.py
janickm: Done -- code and docstrings use `[n, D]` for projection dimension. The implementation was already dimension-agnostic (pure tensor ops).

ncore/impl/sensors/common.py
janickm: Fixed -- consistently uses `t` throughout. The solver derives initial `t` internally from `compute_relative_frame_time(projected_init)`.

ncore/impl/sensors/common.py
janickm: Resolved -- the `stop_error` / `initial_error` parameters no longer exist. The solver uses a single `stop_t_delta` threshold and handles initialization internally.

ncore/impl/sensors/common.py
janickm: Done -- the solver now accepts raw `world_points` and does the dual start/end projection internally. Callers just pass world_points + poses + projector.

ncore/impl/sensors/common.py
janickm: Resolved -- the convergence metric is now purely time-based (`stop_t_delta` = mean absolute change in relative frame time). No projection-space error metric exposed.

ncore/impl/sensors/lidar.py
janickm: Resolved -- the `t_init` / `relative_time` logic is now fully internal to the solver. The solver derives initial times from `projected_init` via `compute_relative_frame_time`, so the comment is no longer relevant.

ncore/impl/sensors/lidar.py
janickm: Resolved -- this code path no longer exists. The solver handles all initialization internally.

ncore/impl/sensors/lidar.py
janickm: Done -- `_LidarRollingShutterProjector` docstring now states: 'Convergence is measured by mean absolute change in relative frame time between iterations.'

ncore/impl/sensors/camera.py
janickm: Done -- `_CameraRollingShutterProjector` docstring now states: 'Convergence is measured by mean absolute change in relative frame time (derived from projected image coordinates) between iterations.'

ncore/impl/sensors/rolling_shutter_test.py
janickm: seems addressed

ncore/impl/sensors/lidar.py
```diff
@@ -470,130 +487,43 @@ def world_points_to_sensor_angles_shutter_pose(
...
-            ),
-            normalized=False,
+        iter_result = RollingShutterSolver.solve(
+            world_points=world_points,
+            T_world_sensor_start=T_world_sensor_start,
+            T_world_sensor_end=T_world_sensor_end,
+            max_iterations=max_iterations,
+            stop_t_delta=stop_t_delta,
+            projector=_LidarRollingShutterProjector(self),
+            stop_t_delta_rate=1e-6,
```
janickm: this should be a parameter to the method, not hardcoded (use this value as default)

ncore/impl/sensors/lidar.py
```diff
@@ -436,8 +454,7 @@ def world_points_to_sensor_angles_shutter_pose(
         start_timestamp_us: Optional[int] = None,
         end_timestamp_us: Optional[int] = None,
         max_iterations: int = 10,
-        stop_mean_relative_time_error: float = 1e-4,
-        stop_delta_mean_relative_time_error: float = 1e-6,
+        stop_t_delta: float = 1e-4,
         return_T_world_sensors: bool = False,
         return_valid_indices: bool = False,
         return_timestamps: bool = False,
```
janickm, line 461: for consistency with camera, let's also add a return_all_projections: bool = False argument (+ add/extend unit tests)

janickm:
/ok to test https://github.com/NVIDIA/ncore/commit/e5dfd7f1b4f7805b0dcbaecc7663443071286e9e

janickm:
Addressed both review comments in 53181896:

1. **`stop_t_delta_rate` is now a parameter** (default `1e-6`) on the lidar method signature, not hardcoded in the solver call. Callers can tune or disable (`0.0`).

2. **Added `return_all_projections: bool = False`** to the lidar method for consistency with the camera API. When enabled, returns sensor angles for all N input points (initial projections for invalid, refined for valid).

/ok to test 53181896

janickm:
## Convergence Analysis

### 1. Camera vs Original: Identical Convergence

camera_new_vs_original

Per-iteration t_delta traces are byte-for-byte identical between new solver and original across all 5 AV driving scenarios. Both converge in 1-3 iterations.

### 2. Lidar vs Original: Identical with Rate-Stop

lidar_new_vs_original


Same convergence path. Green markers show where `stop_t_delta_rate=1e-6` triggers early stopping -- at the exact same iteration as the original solver.

### 3. Lidar Real Data: Plateau from Column Quantization

lidar_convergence_real


Left: t_delta plateaus after 2-3 iterations due to discrete 3600-column quantization (never reaches 1e-4 threshold).
Right: Rate of change `|t_delta[k] - t_delta[k-1]|` drops below 1e-6 at iteration 3-5, triggering the delta-rate early stop.

### 4. Iteration Count Summary

iterations_comparison


Bar chart confirming identical iteration counts between new and original for all AV scenarios (camera: 1-3, lidar: 5-6).

### 5. g(t) Fixed-Point Map (scaled Y-axis)

g_function_scaled


The fixed-point map g(t) with Y-axis zoomed to its actual range. Shows near-zero slope (`|g'(t*)| = 0.0001 to 0.02`). Red cobweb = Picard iteration steps converging to the fixed point.

### 6. Residual f(t) = g(t) - t

g_residual


The residual is nearly linear with slope -1 (since `g' ~ 0`, `f' = g' - 1 ~ -1`). The unique zero crossing is the fixed point. Red dots = iteration points.

### 7. g(t) with Linear Slope Estimate from Endpoints

slope_estimate


Orange dashed line = linear fit using `g(0)` and `g(1)` (which we already compute at initialization). Tracks the staircase closely, confirming the slope is constant and estimable from the two endpoint projections.

### 8. Residual with Newton Step from Slope Estimate

slope_residual


Shows that Newton's method with the endpoint-estimated slope produces the same update as Picard: since `slope ~ 0`, the Newton denominator `(g' - 1) ~ -1`, giving `t_{k+1} = t_k - f/(−1) = t_k + (g - t) = g(t_k)` = Picard.

---

### Key Insight

For standard rolling-shutter sensors with quantized readout (floor-based row/column time), `|g'| << 1` everywhere. **Picard IS Newton in this regime** -- they produce identical updates. The acceleration methods (Secant, Aitken) are reserved for future continuous-time sensors (event cameras, undistorted rolling-shutter with non-linear time progression) where `g'` could approach 1.

PR #120, opened 2026-05-15
fix(types): correct dtype validation in BivariateWindshieldModelParameters

janickm:
## Summary

- Fix copy-paste bug in `BivariateWindshieldModelParameters.__post_init__()` where four assertions all checked `self.horizontal_poly.dtype` instead of the correct field
- `vertical_poly.dtype`, `horizontal_poly_inverse.dtype`, and `vertical_poly_inverse.dtype` are now validated correctly
- Without this fix, invalid data passes validation silently

Replaces #112 which was accidentally auto-merged during a branch mix-up.


janickm:
Already merged via #112.

PR #140, opened 2026-05-27
fix(sensors): respect monotonicity in OpenCVFisheyeCameraModel.compute_max_angle

janickm:
## Problem

`OpenCVFisheyeCameraModel.compute_max_angle` used Newton-Raphson to invert the forward polynomial to the farthest image corner distance, without checking if the polynomial is monotonically increasing up to that point.

For cameras where the image corners are outside the valid FOV (e.g. `f~449` on `1920x1536` with `k=[0.145, -0.067, 0.020, -0.002]`), the polynomial folds before reaching the corner distance, and the Newton solver lands on a nonsensical root:

```python
max_angle = OpenCVFisheyeCameraModel.compute_max_angle(
    np.array([1920, 1536], dtype=np.uint64),
    np.array([449.191, 448.985], dtype=np.float32),
    np.array([959.917, 767.244], dtype=np.float32),
    np.array([0.145375, -0.0673299, 0.0201295, -0.00245998], dtype=np.float32))
# Before: 7.87 rad (451 degrees -- nonsense)
# After:  2.11 rad (121 degrees -- correct)
```

## Solution

Replace the Newton-Raphson inversion with a sampling-based approac that checks two stopping conditions:

1. **Monotonicity**: `dr/dtheta <= 0` (polynomial has folded)
2. **Image boundary**: `r(theta) > max_radius` (projection left the image)

The logic is extracted into a generic `compute_max_angle_with_monotonicity` helper in `common.py` that:
- Reuses the existing `eval_poly_horner` (torch-based, no new dependencies)
- Takes standard polynomial coefficients in Horner form
- Can be reused for `FThetaCameraModel.compute_max_angle` in the future

## BREAKING CHANGE
`OpenCVFisheyeCameraModel.compute_max_angle` is removed. Callers must migrate to the equivalent static method on the parameter dataclass:

```python
  # Before
  from ncore.impl.sensors.camera import OpenCVFisheyeCameraModel
  angle = OpenCVFisheyeCameraModel.compute_max_angle(res, fl, pp, k)

  # After
  from ncore.impl.data.types import OpenCVFisheyeCameraModelParameters
  angle = OpenCVFisheyeCameraModelParameters.compute_max_angle(res, fl, pp, k)
```

## Changes

| File | Change |
|------|--------|
| `ncore/impl/sensors/common.py` | Added `compute_max_angle_with_monotonicity` helper |
| `ncore/impl/sensors/camera.py` | Rewrote `OpenCVFisheyeCameraModel.compute_max_angle` to use the helper |
| `ncore/impl/sensors/camera_test.py` | Relaxed 2 existing test tolerances (sampling precision ~0.0016 rad), added 8 new tests |

## Testing

- All 111 camera tests pass on both Python 3.8 and 3.11
- New tests verify monotonicity guarantee for multiple camera parameter sets
- Specific regression test for the reported problematic intrinsics


janickm:
should be ready for review, waiting for Tyler's comments

Tyler-D:
LGTM! Thanks~

PR #143, opened 2026-06-09
feat(sensors): ideal pinhole camera model and Rectificator

janickm:
## Summary

Adds a distortion-free **ideal pinhole camera model** and a **`Rectificator`** for distorted -> rectified image remapping.

Motivation: a "perfect" pinhole (OpenCV pinhole with zero distortion) is needed both as an efficient camera model and as the natural target for rectification.

## Ideal pinhole camera model

- New abstract `PinholeCameraModelParameters` base sharing principal point, focal length, and the rescale/crop `transform()`. Concrete **siblings**: new `IdealPinholeCameraModelParameters` (type `ideal-pinhole`) and the existing `OpenCVPinholeCameraModelParameters`. An OpenCV pinhole is **not** an instance of an ideal pinhole.
- New abstract `PinholeCameraModel` with closed-form project/unproject reused by `IdealPinholeCameraModel` and `OpenCVPinholeCameraModel`; the latter gains an internal `is_distortion_free` fast-path that skips distortion computation and iterative undistortion.
- Dispatch, serialization (`ideal-pinhole`) and package exports wired.

## Converting any camera to an ideal pinhole

A centralized factory `IdealPinholeCameraModelParameters.from_source(source, target_fov=None)` builds an ideal pinhole approximating any supported source camera. It extracts the source's paraxial pinhole geometry, **preserving the source focal aspect ratio**:

- Ideal / OpenCV pinhole / OpenCV fisheye: the model's own `focal_length` (anisotropic `fx != fy` preserved; OpenCV pinhole distortion dropped).
- FTheta: `[c1 * c, c1]` with `c1 = angle_to_pixeldist_poly[1]` and `c = linear_cde[0]` (since `d(delta)/d(theta)|_0 = c1` and `r = f*tan(theta) ~= f*theta` near 0; shear terms are not representable by a pinhole and are dropped).

`target_fov` (full FOV [rad]) selects the angular extent, **both wider and narrower**:

- `None`: the source's natural FOV. A pinhole focal always yields a FOV < 180deg, so for wide fisheye / omnidirectional cameras this is a narrow rectilinear central window.
- `float`: an isotropic target FOV, preserving the source focal aspect ratio (the binding axis lands exactly at this value).
- `np.ndarray` `[2,]`: per-axis target FOV.

`IdealPinholeCameraModelParameters.natural_fov(source)` (and `fov()` on a built model) return the per-axis FOV so callers can pick / scale a `target_fov` (e.g. `1.2 * natural_fov(source)` to widen). Widening past the captured FOV yields invalid (black-border) regions; `from_source` raises `ValueError` for FOV >= 180deg and `TypeError` for unsupported source models.

## Rectificator

- `Rectificator(source, target)` maps imagery from a source camera into a target camera's image domain (a generic from-camera -> to-camera intrinsic remap). The target may itself be distorted.
- Builds a backward sample map (target -> source) once; exposes `sample_map` and `valid_mask` for custom / GPU / OpenCV resampling.
- `apply(image, mode, padding_mode)` resamples source images (`HW`, `HWC` or `NHWC`) via `grid_sample`, zeroing invalid pixels.
- `source_points_to_target` / `target_points_to_source` map sparse image points between the two cameras.

## Tooling

- `tools/ncore_export_camera.py` gains a `--rectify` mode that rectifies frames to an ideal pinhole via the `Rectificator` before exporting (target derived with `IdealPinholeCameraModelParameters.from_source`):
  - `--rectify-target-fov-deg`: optional target full FOV [deg] of the rectified pinhole (wider or narrower than the natural FOV; omitted -> natural FOV).
  - `--rectify-fov-factor`: multiplicative factor on the (target or natural) FOV (`> 1` widens, `< 1` narrows).
  - Rectified frames are re-encoded (PNG) and the rectified ideal-pinhole intrinsics are written as `<camera_id>.rectified_intrinsics.json`.

## Examples

> Note: images below illustrate earlier `--rectify-max-fov-deg` runs; the flag is now `--rectify-target-fov-deg` (+ `--rectify-fov-factor`) and a wide fisheye's default rectification now succeeds as a narrow rectilinear central window (rather than failing).

### Perspective lens

Original frame
000000_original

Rectified (natural FOV)
000000_rectified

Rectified (target FOV=60deg)
000000_targetfov_60

Rectified (target FOV=90deg)
000000_targetfov_90

Rectified (FOV factor 0.7)
000000_factor_0 7

Rectified (FOV factor 1.3)
000000_factor_1 3

### Fisheye lens

Original frame
000000_original

Rectified (default FOV)
000000_rectified_default

Rectified (target FOV=90deg)
000000_targetfov_90

Rectified (target FOV=120deg)
000000_targetfov_120

Rectified (target FOV=150deg)
000000_targetfov_150

Rectified (FOV factor 0.5)
000000_factor_0 5

## Breaking change

The already-rectified dataset converters (**KITTI**, **nuScenes**) now emit the ideal pinhole model instead of a zero-distortion OpenCV pinhole. Downstream consumers expecting `opencv-pinhole` from these converters must update. Existing stored data still decodes unchanged (the `opencv-pinhole` / `pinhole` decode paths are untouched).

## Testing

- Ideal pinhole model tests (dispatch, round-trip, parity vs zero-coeff OpenCV pinhole, `is_distortion_free`, jacobian path), serialization, and `transform`.
- `IdealPinholeCameraModelParameters.from_source` / `natural_fov` / `fov` tests: default per source model (incl. FTheta `c`-fold and anisotropic OpenCV focal preserved), scalar (aspect-preserving) and per-axis `target_fov`, narrow/widen, `natural_fov` round-trip, feasibility and unsupported-type errors.
- `rectification_test.py` (map shapes, map vs explicit projection, `apply` across image layouts/dtypes, invalid-pixel zeroing, identity rectification, point round-trip, pure/narrow/widen targets).
- Full `camera_test.py` + `rectification_test.py`: 97 passed, 1 skipped (GPU), 12156 subtests passed. Existing tests unbroken.
- `bazel run format` clean; ruff + ty clean on changed files.


docs/data/sensor_models.rst
```diff
@@ -312,6 +359,92 @@ The following parameters will be available in ``external_distortion_parameters``
...
+1. each target pixel is unprojected to a ray with the target model,
+2. the ray is re-projected to a source image coordinate with the source model.
+
+It exposes:
+
+* ``sample_map`` - ``[H, W, 2]`` source image coordinates to sample for each
+  target pixel (usable directly with custom / GPU / OpenCV resampling).
+* ``valid_mask`` - ``[H, W]`` boolean mask of target pixels whose ray projects
+  inside the source image.
+* ``apply(image, mode="bilinear", padding_mode="zeros")`` - resamples a source
```
janickm, line 451: we should shortly outline the possible modes / padding variants

docs/data/sensor_models.rst
```diff
@@ -312,6 +359,92 @@ The following parameters will be available in ``external_distortion_parameters``
...
+  image (``[H, W]``, ``[H, W, C]`` or ``[N, H, W, C]``, channels-last) into the
+  target domain, zeroing invalid pixels.
+* ``source_points_to_target(points)`` / ``target_points_to_source(points)`` -
+  map sparse continuous image points between the two cameras.
+
+.. note::
+
+   The map only encodes an intrinsic remap within a shared frame. Stereo
```
janickm: please remove this note

ncore/impl/data/types.py
```diff
@@ -364,30 +393,102 @@ def transform(
     PolynomialType = FThetaCameraModelParameters.PolynomialType
 
 
+def _max_corner_distance(resolution: np.ndarray, principal_point: np.ndarray) -> Tuple[float, float]:
```
janickm: I don't like these free helpers - either put them into local scope of the IdealPinholeParameters (or base class) pack or to some utility module maybe?

ncore/impl/data/types.py
```diff
@@ -364,30 +393,102 @@ def transform(
...
+    """Largest half-extent from the principal point to an image border along each axis [px]"""
+    half_extent_x = max(float(principal_point[0]), float(resolution[0]) - float(principal_point[0]))
+    half_extent_y = max(float(principal_point[1]), float(resolution[1]) - float(principal_point[1]))
+    return half_extent_x, half_extent_y
+
+
+def _assert_pinhole_fov_feasible(
```
janickm: feels pinhole specific, should be a static method of the IdealPinholeParameters

ncore/impl/data/types.py
```diff
@@ -364,30 +393,102 @@ def transform(
...
+    fov_y = 2.0 * math.atan2(half_extent_y, float(focal_length[1]))
+
+    if max(fov_x, fov_y) > max_fov:
+        raise ValueError(
+            f"Required pinhole field of view ({math.degrees(max(fov_x, fov_y)):.1f} deg) "
+            f"exceeds max_fov ({math.degrees(max_fov):.1f} deg)"
+        )
+
+
+def _fit_ideal_focal(
```
janickm: same

ncore/impl/data/types.py
```diff
@@ -364,30 +393,102 @@ def transform(
...
+    return np.array(
+        [max(float(focal_length[0]), min_focal_x), max(float(focal_length[1]), min_focal_y)],
+        dtype=np.float32,
+    )
+
+
 @dataclass
-class OpenCVPinholeCameraModelParameters(CameraModelParameters, dataclasses_json.DataClassJsonMixin):
-    """Represents Pinhole-specific (OpenCV-like) camera model parameters"""
+class PinholeCameraModelParameters(CameraModelParameters):
```
janickm, line 369: should this be a dataclasses_json.DataClassJsonMixin as it's "final"?

janickm: Let's make sure we have a unit test to test re-serialization of these

ncore/impl/data/types.py
```diff
@@ -453,6 +545,103 @@ def transform(
             focal_length=self.focal_length * image_domain_scale_factors,
         )
 
+    def to_ideal(self, max_fov: Optional[float] = None) -> IdealPinholeCameraModelParameters:
```
janickm: do we need this - should only be required in non-abstract / final classes?

ncore/impl/sensors/camera.py
```diff
@@ -1433,17 +1547,23 @@ def __init__(
...
         assert self.tangential_coeffs.shape == (2,)
         assert self.tangential_coeffs.dtype == self.dtype
         assert self.thin_prism_coeffs.shape == (4,)
         assert self.thin_prism_coeffs.dtype == self.dtype
 
+        # Whether the model is free of any (lens) distortion - allows skipping the
+        # distortion computation and iterative undistortion and falling back to the
+        # closed-form ideal pinhole path. External distortion is handled separately by
+        # the base CameraModel and does not affect this flag.
+        self._is_distortion_free: bool = bool(
```
janickm: can you check via the existing check in the passed camera_model_parameters ?

ncore/impl/sensors/camera_test.py
```diff
@@ -1897,5 +1899,213 @@ def test_monotonicity_guarantee_multiple_cameras(self):
...
+        # focal length is the first-order forward polynomial coefficient
+        self.assertAlmostEqual(float(ideal.focal_length[0]), 1250.0, places=3)
+        self.assertAlmostEqual(float(ideal.focal_length[1]), 1250.0, places=3)
+
+
+class TestIdealPinholeParameterIO(unittest.TestCase):
+    """Serialization round-trip for the ideal pinhole parameters"""
+
+    def test_encode_decode(self):
+        from ncore.impl.data.types import decode_camera_model_parameters, encode_camera_model_parameters
```
janickm: don't use any "local" imports anywhere

ncore/impl/sensors/camera_test.py
```diff
@@ -1897,5 +1899,213 @@ def test_monotonicity_guarantee_multiple_cameras(self):
...
+        ideal = params.to_ideal()
+        # focal length is the first-order forward polynomial coefficient
+        self.assertAlmostEqual(float(ideal.focal_length[0]), 1250.0, places=3)
+        self.assertAlmostEqual(float(ideal.focal_length[1]), 1250.0, places=3)
+
+
+class TestIdealPinholeParameterIO(unittest.TestCase):
+    """Serialization round-trip for the ideal pinhole parameters"""
+
+    def test_encode_decode(self):
```
janickm, line 2289: do we have round-trip tests for the opencv pinhole one also (maybe in an existing test on intrinsicre- serializations?), as that one was adapted also?

ncore/impl/sensors/rectification.py
```diff
@@ -0,0 +1,201 @@
...
+        """Boolean mask of valid target pixels, shape ``[H, W]``
+
+        A target pixel is valid when its ray projects to a point inside the source image.
+        """
+        return self._valid_mask
+
+    def apply(
+        self,
+        image: Union[torch.Tensor, np.ndarray],
+        mode: str = "bilinear",
```
janickm: use Literal[] s with the possible values for both arguments

ncore/impl/sensors/rectification.py
```diff
@@ -0,0 +1,201 @@
...
+    ) -> torch.Tensor:
+        """Resample a source image into the target camera domain
+
+        Args:
+            image: source image, either ``[H_src, W_src]``, ``[H_src, W_src, C]`` or
+                   ``[N, H_src, W_src, C]`` (channels-last, as commonly stored). Integer
+                   images are converted to float for interpolation.
+            mode: interpolation mode passed to ``grid_sample`` (``"bilinear"`` or
+                  ``"nearest"``).
+            padding_mode: out-of-bounds sampling mode passed to ``grid_sample``.
```
janickm: let's list all possible values (also in the literal argument)

ncore/impl/sensors/rectification.py
```diff
@@ -0,0 +1,201 @@
...
+            mode=mode,
+            padding_mode=padding_mode,
+            align_corners=False,
+        )  # [N, C, H_tgt, W_tgt]
+
+        # Back to channels-last and zero out invalid target pixels
+        result = sampled.permute(0, 2, 3, 1)
+        result = result * self._valid_mask[None, :, :, None].to(result.dtype)
+
+        if not was_floating:
```
janickm, line 164: do we have unit tests for both floating / integer cases?

ncore/impl/sensors/rectification_test.py
```diff
@@ -0,0 +1,167 @@
...
+    import os
+
+    if os.environ.get("NCORE_NO_GPU_TESTS"):
+        return (torch.device("cpu"),)
+    if torch.cuda.is_available():
+        return (torch.device("cpu"), torch.device("cuda"))
+    return (torch.device("cpu"),)
+
+
+def _distorted_pinhole(device, dtype) -> CameraModel:
```
janickm: also test with other distorted sources, like ftheta / opencv-fisheye

ncore/impl/sensors/rectification_test.py
```diff
@@ -0,0 +1,167 @@
...
+        principal_point=np.array([320.0, 240.0], dtype=np.float32),
+        focal_length=np.array([300.0, 300.0], dtype=np.float32),
+        radial_coeffs=np.array([-0.2, 0.05, 0.0, 0.0, 0.0, 0.0], dtype=np.float32),
+        tangential_coeffs=np.zeros(2, dtype=np.float32),
+        thin_prism_coeffs=np.zeros(4, dtype=np.float32),
+    )
+    return CameraModel.from_parameters(params, device=device, dtype=dtype)
+
+
+def _ideal_target(device, dtype) -> CameraModel:
```
janickm: also test differently scaled targets

docs/data/sensor_models.rst
janickm: Done -- the `apply` bullet now lists the interpolation `mode` values (`bilinear`, `nearest`, `bicubic`) and the `padding_mode` values (`zeros`, `border`, `reflection`), noting both are forwarded to `grid_sample`.

docs/data/sensor_models.rst
janickm: Removed. Dropped the stereo note from both the docs and the `Rectificator` class docstring.

ncore/impl/data/types.py
janickm: Done -- removed the free helpers. They are now static methods on `IdealPinholeCameraModelParameters` (`_max_corner_half_extent`, `_assert_fov_feasible`, `_fit_focal`), since they are pinhole/ideal-specific.

ncore/impl/data/types.py
janickm: Done -- now `IdealPinholeCameraModelParameters._assert_fov_feasible` (static method).

ncore/impl/data/types.py
janickm: Done -- now `IdealPinholeCameraModelParameters._fit_focal` (static method).

ncore/impl/data/types.py
janickm: It should not -- `PinholeCameraModelParameters` is the *abstract* base (it has no `type()` and is not instantiable), so it intentionally does not mix in `DataClassJsonMixin`. The two concrete/final subclasses (`IdealPinholeCameraModelParameters`, `OpenCVPinholeCameraModelParameters`) already are `DataClassJsonMixin`.

janickm: Done -- added `TestParameterIO.test_encode_decode_roundtrip` which round-trips `encode`/`decode_camera_model_parameters` for every model type (ftheta, ideal-pinhole, opencv-pinhole, opencv-fisheye, plus windshield variants), and added an ideal-pinhole entry to the shared `cam_model_params` list.

ncore/impl/data/types.py
janickm: Right -- removed it from the base. `to_ideal` is now an `@abstractmethod` on `CameraModelParameters`, and only the final classes implement it (ideal, opencv-pinhole, opencv-fisheye, ftheta). The abstract `PinholeCameraModelParameters` no longer carries a stub.

ncore/impl/sensors/camera.py
janickm: Done -- `OpenCVPinholeCameraModel.__init__` now sets `self._is_distortion_free = camera_model_parameters.is_distortion_free`, reusing the parameters' own check (which also accounts for external distortion).

ncore/impl/sensors/camera_test.py
janickm: Done -- moved `encode_camera_model_parameters` / `decode_camera_model_parameters` to the top-level imports; no local imports remain (also removed the local `import os` in the rectification test helper).

ncore/impl/sensors/camera_test.py
janickm: Yes -- `TestParameterIO.test_encode_decode_roundtrip` now exercises the opencv-pinhole serialization too (the whole `cam_model_params` list, which includes the distorted opencv-pinhole entry), so the adapted class is covered.

ncore/impl/sensors/rectification.py
janickm: Done -- `mode: Literal["bilinear", "nearest", "bicubic"]` and `padding_mode: Literal["zeros", "border", "reflection"]`.

ncore/impl/sensors/rectification.py
janickm: Done -- both the `Literal` args and the docstring now list all possible values.

ncore/impl/sensors/rectification.py
janickm: Yes -- `test_apply_float_and_integer` now covers float32, uint8 and int32 inputs (and asserts integer dtypes are preserved), plus grayscale and batched NHWC.

ncore/impl/sensors/rectification_test.py
janickm: Done -- added ftheta and opencv-fisheye source factories; `test_map_matches_explicit_projection` and `test_point_roundtrip` are now parameterized over all three distorted source models.

ncore/impl/sensors/rectification_test.py
janickm: Done -- added `test_scaled_target` (zoom in 1.5x and out 0.6x via `to_ideal().transform(...)` with new resolution), validating map/mask/output shapes and that the map still matches the explicit projection at the scaled-target center.

ncore/impl/data/types.py
```diff
@@ -454,6 +487,170 @@ def transform(
...
+    ) -> np.ndarray:
+        """Derive an ideal-pinhole focal length from an approximate focal, honouring FOV limits
+
+        Starting from an approximate ``focal_length``, the focal is increased (zoomed in)
+        if necessary so that the resulting pinhole half field of view to the farthest
+        image border does not exceed ``source_max_angle`` (the captured field of view; a
+        pinhole cannot represent rays beyond it) nor, if provided, ``max_fov / 2``.
+
+        Args:
+            source_max_angle: the source camera's ``max_angle`` half-FOV [rad].
```
janickm: not all args are listed

ncore/impl/sensors/rectification_test.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+
+
+def _ftheta_params() -> FThetaCameraModelParameters:
+    return FThetaCameraModelParameters(
+        resolution=np.array([640, 480], dtype=np.uint64),
+        shutter_type=ShutterType.GLOBAL,
+        principal_point=np.array([320.0, 240.0], dtype=np.float32),
+        reference_poly=FThetaCameraModelParameters.PolynomialType.ANGLE_TO_PIXELDIST,
+        pixeldist_to_angle_poly=np.array([0.0, 0.002, 0.0, 0.0, 0.0, 0.0], dtype=np.float32),
```
janickm: please use one _with_ distortion

ncore/impl/data/types.py
janickm: Done -- the `_fit_focal` docstring now lists all args (`resolution`, `focal_length`, `principal_point`, `source_max_angle`, `max_fov`).

ncore/impl/sensors/rectification_test.py
janickm: Done -- the ftheta source factory now uses a real, genuinely distorted (nonlinear) bw-based F-Theta model (the validated fixture from `camera_test.py`) instead of the previous linear-only polynomials.

PR #128, opened 2026-05-21
feat(nuscenes/lidar-model): add nuScenes V4 converter and lidar-model estimation lib / evaluation tool

janickm:
## Summary

Add a nuScenes dataset converter for NCore V4 format, a generic structured lidar model extraction library, and a lidar model evaluation tool.

image

Closes #123

## Architecture

The work is split into three self-contained commits:

### 1. `feat(data_converter): add structured lidar model extraction library`

`tools/data_converter/structured_lidar_model.py` -- a generic, reusable library for deriving structured spinning lidar models from motion-compensated point cloud data. Designed for datasets that lack raw sensor timestamps.

**Key features:**
- Composable steps: `extract_column_azimuths`, `compute_column_alignment`, `assign_model_columns`, `compute_frame_timestamps`, `decompensate_frame`
- Resolution upsampling (`upsample_model`): interpolate column azimuths to 2x/4x for sub-column alignment precision
- Multi-frame optimization (`optimize_model`): median-based correction across many frames
- Pre-computed timestamps support: `align_frame(timestamps_us=...)` for sensors with known per-point timestamps
- HDL-32E presets: spec elevations, analytical firing offsets, nominal model factory
- 18 unit tests (no external data dependency)

### 2. `feat(tools): add lidar model evaluation tool`

`tools/ncore_evaluate_lidar_model.py` -- CLI tool that evaluates model quality by comparing model-predicted vs stored native directions. Works on any V4 sequence.

**Outputs:**
- Numerical metrics: angular error (mean/median/p95), per-row and per-frame breakdowns, pixel-equivalent error
- Overlay images: native (cyan) + model (red) points on camera
- Error heatmap images: points colored by angular error magnitude

000070_overlay

### 3. `feat(nuscenes): add nuScenes to NCore V4 converter`

Full converter handling 6 cameras, 1 lidar (HDL-32E), 5 radars, 3D cuboid annotations. Uses the generic library directly (no local lidar model code).

## Lidar Model Quality

| Dataset | Mean error (far) | Pixel equiv | Notes |
|---------|-----------------|-------------|-------|
| Waymo (ground truth) | 0.010 deg | 0.35 px | Theoretical floor |
| **PAI (extraction)** | **0.028 deg** | ~0.5 px | Factory-quality |
| **nuScenes (4x nominal + opt)** | **0.029 deg** | 0.7 px | **This PR** |
| nuScenes (1x empirical + opt) | 0.083 deg | 1.8 px | Without resolution upsampling |

[...]

PR #145, opened 2026-06-10
feat(argoverse2): add Argoverse 2 Sensor Dataset to NCore V4 converter

### LINKED ISSUE #123: nuScenes and Argoverse 2 support
### Problem Statement

How to convert from nuScenes / Argoverse 2 to Ncore format? 

### Proposed Solution

Hi,

This is a great initiative. It's also awesome to see new features like the KITTI converter get rolled out.

Are other datasets support also in the works currently?
(Specifically, **nuScenes** and **Argoverse 2** converters would be very useful for me.)

Is there a roadmap for converters and other features rollout that can be tracked?

Thanks,
Samin

### Alternatives Considered

_No response_

### Additional Context

_No response_

janickm:
## Summary

Add a converter for the **Argoverse 2 (AV2) Sensor Dataset** to NCore V4, closing the Argoverse 2 half of #123 (the nuScenes half landed in #128).

image

The converter reads the AV2 on-disk Apache Feather files directly with `pyarrow` (Arrow tables, no pandas) and deliberately avoids the heavy `av2` devkit (which pulls in torch, kornia, numba, polars, PyAV). The only added dependency is `pyquaternion`, already used in this package.

Closes #123

## Sensors

- **Cameras**: all 9 global-shutter cameras (7 ring + 2 stereo). AV2 imagery is shipped already undistorted -- the official av2 devkit projects with the intrinsic matrix `K` only and does not load the `k1, k2, k3` columns -- so the stored model is an ideal (distortion-free) pinhole (`IdealPinholeCameraModelParameters`, `ShutterType.GLOBAL`). The `k1, k2, k3` in `intrinsics.feather` describe the original lens (for re-distorting into the raw frame) and are intentionally not applied. Note: the stereo pair is genuinely monochrome (R=G=B in the source JPEGs).
- **Lidar**: the two stacked Velodyne VLP-32C units (`up_lidar` / `down_lidar`) are stored separately, each with its own extrinsic. AV2 sweeps are egomotion-compensated to the sweep reference timestamp (the sweep start) and expressed in the egovehicle frame, with real per-point timestamps (`offset_ns`). Points are mapped into each unit's sensor frame and decompensated -- referenced to the sweep start -- using the shared `MotionCompensator` `reference_timestamp_us`, so NCore stores raw per-point-time directions.
- **Radar**: AV2 has none.
- **Cuboids**: native to the egovehicle frame at the sweep time; stored in the `rig` frame at that timestamp with no ego pose baked in, so egomotion stays swappable downstream (a V4 feature).

## Lidar unit split

AV2 shares one `laser_number` range `[0, 63]` across both units with no documented up/down mapping. The two halves (`< 32` / `>= 32`) are the two physical units; the up/down label is recovered from extrinsic geometry by per-beam elevation flatness (a laser ring traces a constant-elevation cone only in its own sensor frame), decided once per log with a wide, stable margin.

## Structured VLP-32C lidar model

AV2 provides no native firing-column index, but `offset_ns` + `laser_number` reconstruct the firing pattern (`offset_ns` -> firing column at 10 Hz, `laser_number` -> beam/row). A structured model is derived per unit and stored as intrinsics with per-point `model_element`. Reaching native-sensor accuracy required several steps, each found by evaluating with `ncore_evaluate_lidar_model` across many logs:

1. **Decompensate first.** The geometry is derived from the *decompensated* reference sweep; the raw motion-compensated cloud is azimuth-smeared by ego motion (~0.5 deg).
2. **Empirical per-row azimuth offsets.** The 32 beams of a firing column span ~8.5 deg of azimuth (not co-azimuthal), so per-row offsets are fit empirically (jointly with per-column azimuths), not assumed from an HDL-32E pattern.
3. **Per-unit spin direction.** The two stacked units fire in opposite phase and so spin oppositely in their own frames (one `cw`, one `ccw`); detected from the data.
4. **One-revolution column wrap.** An AV2 sweep covers ~1.02 revolutions, so columns are sized to one revolution and wrapped modulo, keeping the azimuth ramp below 2*pi.
5. **Per-frame affine alignment.** The model is static, but the spin phase at a given `offset_ns` drifts ~1 deg between sweeps and the spin rate drifts slightly within a sweep on some scenes. Each frame is re-aligned by an affine column remap (constant phase + linear term).
6. **4x column upsampling** so the per-frame alignment is not column-quantized.
7. **Near-range offset fallback** for steep downward beams that only return at near range (e.g. the lowest laser at ~-25 deg, with no far data).

A `--lidar-model-source {empirical,none}` flag (default `empirical`) gates emission.

### Validation

Converted **38 val logs (76 lidar units)** and evaluated each with `ncore_evaluate_lidar_model`:

| Metric | mean | p90 | max |
|---|---|---|---|
| median far-range error | 0.033 deg | 0.055 deg | 0.077 deg |
| azimuth bias (signed) | 0.005 deg | -- | 0.022 deg |
| elevation bias (signed) | 0.005 deg | -- | 0.026 deg |

All 76 units are sub-0.08 deg median with **no systematic azimuth or elevation bias** -- on par with native-column sensors. (Isolated per-point maxima of 1-16 deg are dynamic objects / steepest grazing-beam returns, which no static spin model can represent; per-frame p95 stays ~0.08 deg.)

## Coordinate frames

The first ego pose's `city_SE3_egovehicle` is stored as the static `world -> world_global` anchor, so `world_global` is the AV2 city frame and absolute coordinates stay recoverable for later HD-map alignment.

## Testing

```bash
# Data-free unit tests (run in CI)
bazel test //tools/data_converter/argoverse2:pytest_utils

# Integration test (requires a downloaded AV2 log)
AV2_DIR=/path/to/argoverse2/sensor AV2_SPLIT=val \
    bazel test //tools/data_converter/argoverse2:pytest_converter
```

- `pytest_utils` is a **data-free** unit test for the VLP-32C model derivation: reconstruction accuracy on a synthetic sweep, the >1-revolution column wrap, cross-frame phase generalization, and intra-sweep rate drift. Each guard was verified to fail when its fix is reverted.
- `pytest_converter` (manual / `AV2_DIR`-gated) validates the full pipeline against a real `val` log, including a lidar-vs-cuboid alignment check and the model-reconstruction accuracy guard.

Builds on the merged `MotionCompensator.reference_timestamp_us` / `anchor_frame_id` work (#146, #147) and the lidar-model robustness fixes (#148).

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -1502,7 +1502,7 @@
         "bzlTransitiveDigest": "WyF2tLBpR8cyete55VmU+AwLC/mhDbnIlEdbVymbm04=",
         "usagesDigest": "DXk0VBJUK7HdLXJE59BVPG4lCTTPnHGNQg/SH9dTuVE=",
         "recordedFileInputs": {
-          "@@//deps/pip/requirements_3_11.txt": "01687dd0690641cecfe0cfa3a5a8ba2186f65112c34b4ab7b64caa55a416ae56",
+          "@@//deps/pip/requirements_3_11.txt": "845ddefb3a79c099a05f39880752241d3c30cefb3774c0611f253d5bff3f774f",
           "@@//deps/pip/requirements_3_8.txt": "94c254d645dc1d0555f48e66fa499aa2efecfc5e547500cd4ed409aee28e8880",
           "@@rules_fuzzing+//fuzzing/requirements.txt": "ab04664be026b632a0d2a2446c4f65982b7654f5b6851d2f9d399a19b7242a5b",
           "@@rules_python+//tools/publish/requirements_darwin.txt": "095d4a4f3d639dce831cd493367631cd51b53665292ab20194bac2c0c6458fa8",
--- a/deps/pip/BUILD.bazel
+++ b/deps/pip/BUILD.bazel
@@ -34,6 +34,7 @@ pip_compile(
     size = "medium",
     args = PIP_COMPILE_ARGS,
     data = [
+        ":requirements_argoverse2.in",
         ":requirements_colmap.in",
         ":requirements_docs.in",
         ":requirements_ncore.in",
--- a/deps/pip/requirements_3_11.txt
+++ b/deps/pip/requirements_3_11.txt
@@ -1853,7 +1853,9 @@ pyarrow==23.0.1 \
     --hash=sha256:f4b0dbfa124c0bb161f8b5ebb40f1a680b70279aa0c9901d44a2b5a20806039f \
     --hash=sha256:fa8e51cb04b9f8c9c5ace6bab63af9a1f88d35c0d6cbf53e8c17c098552285e1 \
     --hash=sha256:fed7020203e9ef273360b9e45be52a2a47d3103caf156a30ace5247ffb51bdbd
-    # via -r deps/pip/requirements_pai.in
+    # via
+    #   -r deps/pip/requirements_argoverse2.in
+    #   -r deps/pip/requirements_pai.in
 pycocotools==2.0.11 \
     --hash=sha256:04480330df5013f6edd94891a0ee8294274185f1b5093d1b0f23d51778f0c0e9 \
     --hash=sha256:08c79789fd79e801ae4ecfcfeec32b31e36254e7a2b4019af28c104975d5e730 \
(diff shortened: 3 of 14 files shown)


janickm:
FYI @mpelko-nvidia on the lidar-model derivation for argoverse

PR #150, opened 2026-06-13
Fix IndexedTarStore writer reload truncation

fallintoplace:
## Summary

`IndexedTarStore.reload_resources()` reopened write-mode archives with `wb+`, which truncates the archive during an in-progress write. This changes the writer reload path to reopen the existing archive with `rb+` so already-written tar payloads remain intact.

I also added a regression test that writes one entry, reloads writer resources, verifies the file size and payload are preserved, writes a second entry, and then reopens the archive in read mode to check both entries.

## Test

`PYTHONPATH=$PWD uv run --no-project --python 3.12 --with 'pytest==8.3.4' --with 'cbor2==5.9.0' --with 'numcodecs==0.15.1' --with 'numpy==1.26.4' --with 'parameterized==0.9.0' --with 'universal-pathlib==0.3.6' --with 'zarr==2.18.4' pytest -q ncore/impl/data/stores_test.py`

Result: `26 passed`

### DIFF
--- a/ncore/impl/data/stores.py
+++ b/ncore/impl/data/stores.py
@@ -262,9 +262,9 @@ class IndexedTarStore(Store):
                 self.tar_file_object = self.itar_upath.open("rb")
             else:
                 # universal_path for Python 3.8 (<=0.2.6) doesn't expose a
-                # write/read mode in it's static type-hints, although "wb+" is still accepted
+                # write/read mode in it's static type-hints, although "rb+" is still accepted
                 # if the FS supports it, so ignore type-checker here
-                self.tar_file_object = self.itar_upath.open("wb+")  # type: ignore[call-overload]
+                self.tar_file_object = self.itar_upath.open("rb+")  # type: ignore[call-overload]
 
             if self.mode == "w":
                 # assign the new file object to the tar file (only required for write mode)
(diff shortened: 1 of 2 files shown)


janickm:
This is great, thanks for fixing @fallintoplace - LGTM.

Can you please still correct the commit message to conform to the conventional-commit requirements (maybe use `fix(itar-store): ...` as the commit message prefix). If you can, please also sign the commit according to the [DCO](https://github.com/NVIDIA/ncore/blob/main/CONTRIBUTING.md#commit-signatures).

I think we never ran into this as we only ever reloaded resources when reading (e.g., when training multi-process @ torch dataloader, where the stores are loaded in the main process, requiring a reload in the forked workers) - maybe I ask in which setting you ran into this while writing (wondering how you could have a setting where you'd need to reload the resources while writing)?

janickm (changes requested):
Code looks great - just the commit message requires fixes + signature are missing. Maybe adding details to the commit body would also be useful.

janickm (approved):

janickm:
thanks @fallintoplace! approved, merging once CI succeeds

do you mind still sharing in which types of setups you ran into this problem (reloading resources while writing?)

PR #163, opened 2026-07-16
fix(pai-converter): center ego-vehicle bbox z-gate at height/2

Willyzw:
## Description

Fixes a bug in the PAI converter's ego-vehicle bbox filter that leaves Lidar self-return points near the ego roof unfiltered.

**Root cause.** In `tools/data_converter/pai/utils.py`, the function `filter_ego_vehicle_points` only offsets the box center in x (`rear_axle_to_bbox_center`), while y and z stay centered at 0. According to the [PAI coordinate-system docs](https://github.com/NVlabs/physical_ai_av/wiki/2.--Coordinate-Systems), the rig z origin is at ground level, so the z-gate spans `[-(h/2+pad), h/2+pad]` instead of `[ground-pad, roof+pad]`. Thus the self-return points at upper roof part fall outside the box and are not filtered. 

**Fix.** Center the z-gate at `height/2` so the box spans from ground to roof:

**Verification**. Confirmed on a PAI clip as shown below, red = box before this fix, green = after (rear y-z views); the red box top cuts through the middle of ego vehicle while the green box encloses the vehicle entirely.

ego bbox before/after

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update

### DIFF
--- a/tools/data_converter/pai/utils.py
+++ b/tools/data_converter/pai/utils.py
@@ -56,8 +56,13 @@ def filter_ego_vehicle_points(
     half_width = vehicle_bbox["width"] / 2 + padding
     half_height = vehicle_bbox["height"] / 2 + padding
 
-    # Center offset
+    # Box center in the rig frame. x uses the rear-axle-to-bbox-center offset; z
+    # uses the vehicle mid-height because the rig z origin is at ground level, so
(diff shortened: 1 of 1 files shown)


janickm:
Verified fix locally, thanks for spotting and correcting @Willyzw !

Merging once CI is happy

PR #165, opened 2026-07-21
docs: add LiDAR model estimation tutorial

kangx326:
## What changed

- Add a new LiDAR model estimation tutorial to the Tutorials section.
- Explain nominal and empirical structured spinning LiDAR model construction.
- Document per-frame `model_element` alignment, quality evaluation, and NCore V4 writing.
- Add the local tutorial source to the documentation Bazel target and Sphinx navigation.

## Why

AV dataset integrations need both a structured LiDAR intrinsic model and per-ray `(row, column)` indices. The existing APIs and converter utilities support this workflow, but the steps were spread across source files and tool documentation.

## Impact

Users now have one end-to-end guide, following the existing data-loading tutorial style, for estimating, validating, and storing LiDAR models in NCore V4.

## Validation

- Standalone Sphinx 8.1.3 HTML build with warnings treated as errors (`-W`): passed with zero warnings.
- Parsed all six Python code blocks with `ast.parse`: passed.
- `git diff --check`: passed.
- Full `bazel build //docs:ncore` could not complete locally because the private `ncore-docs-data` package requires GitHub Packages credentials; with a local asset override, macOS arm64 dependency resolution then stopped at the repository's CUDA-only `torch==2.7.0+cu128` pin before Sphinx execution.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -17,7 +17,7 @@ load("//bazel/sphinx:defs.bzl", "sphinx_html_gen", "sphinx_view")
 
 sphinx_html_gen(
     name = "ncore",
-    srcs = glob(["apis/**"]) + glob(["data/*.rst"]) + glob(["tools/*.rst"]) + glob(["conversions/**/*.rst"]) + glob(["_static/**"]) + glob(["_templates/**"]) + [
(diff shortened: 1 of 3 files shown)


janickm:
[sorry, I wasn't following up on this - given we are going towards a model-less setup for lidar in general + I'm not confident in the current version yet I'd postpone this change into the mainline repo for now - the existing samples (waymo / nuscenes) have extensive details on how to derive a lidar model / model elements in specific dataset-dependent settings]

PR #171, opened 2026-08-21
feat(packaging): ship PAI converter in nvidia-ncore wheel

eattia-nvidia:
## Description

Package the PAI data converter in the `nvidia-ncore` wheel so PhysicalAI-AV clips can be converted without a source checkout or Bazel.

- Add `pai_converter_pkg` to `//ncore:ncore_wheel` (converter + `pai_remote` + CLI).
- Expose console script `ncore-convert` → `tools.data_converter.pai.converter:cli`.
- Add optional extra `nvidia-ncore[pai]` for converter deps (Click, DracoPy, imageio, pandas, pyarrow, PyNvVideoCodec, remotezip, requests, tqdm, debugpy).
- Document installed usage in `ncore/PYPI_README.md` and `tools/data_converter/pai/README.md`.
- Cover packaging with `//ncore:pytest_wheel_packaging_3_11`.

`pai-clip-dl` is not shipped as its own console script in this PR; streaming (`pai-stream-v4`) is the primary no-checkout path. Camera decode via PyNvVideoCodec needs an NVIDIA GPU (Turing+).

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Checklist

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] My commits are GPG-signed
- [x] I have added tests that prove my fix is effective or my feature works
- [x] Code is formatted (`bazel run //:format`)
- [x] I have updated documentation as needed
- [x] My changes include SPDX license headers on all new files

### DIFF
--- a/bazel/packaging/namespace_converter_srcs.py
+++ b/bazel/packaging/namespace_converter_srcs.py
@@ -1,0 +1,85 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rewrite in-repo tools.* converter sources into ncore.converters for the wheel."""
+
+from __future__ import annotations
+
+import argparse
+import sys
+
+from pathlib import Path
+from typing import List, Optional
+
+
+_SPDX_INIT = """# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""
+
+_REPLACEMENTS = (
+    ("from tools.data_converter", "from ncore.converters"),
+    ("import tools.data_converter", "import ncore.converters"),
+    ("from tools.debug", "from ncore.converters.debug"),
+    ("from ncore_repo.tools.debug", "from ncore.converters.debug"),
+    ("from external.ncore_repo.tools.debug", "from ncore.converters.debug"),
+)
+
+
+def _dest_for(src: Path) -> Path:
+    parts = src.parts
+    if src.name == "debug.py":
+        return Path("debug.py")
+    if "pai_remote" in parts:
+        return Path("pai") / "pai_remote" / src.name
+    if "pai" in parts:
+        return Path("pai") / src.name
+    if src.name == "cli.py":
+        return Path("cli.py")
+    raise ValueError(f"no wheel path mapping for {src}")
+
+
+def _rewrite(text: str) -> str:
+    for old, new in _REPLACEMENTS:
+        text = text.replace(old, new)
+    return text
+
+
+def main(argv: Optional[List[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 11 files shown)


ncore/BUILD.bazel
```diff
@@ -45,6 +46,25 @@ py_wheel(
     author = "NVIDIA Corporation",
     description_file = "PYPI_README.md",
     distribution = "nvidia-ncore",
+    entry_points = {
+        "console_scripts": [
+            "ncore-convert = tools.data_converter.pai.converter:cli",
+        ],
+    },
+    extra_requires = {
+        "pai": [
```
andmax, line 55: The pai extra is missing runtime dependencies, so pip install "nvidia-ncore[pai]" gives an import error rather than a working converter. tools/data_converter/pai/pai_remote/downloader.py imports rich, which isn't listed,  installing the extra and importing that module fails with ModuleNotFoundError: No module named 'rich'. Comparing against the dependency set needed to build the converter as a bazel target, scipy and universal-pathlib also look absent. Worth being clear this isn't introduced here: rich is undeclared in every deps/pip/requirements_*.in on main today. It's just that this MR is what makes the module installable by external users, so a latent gap becomes a user-visible break at exactly the moment the extra ships.

ncore/BUILD.bazel
```diff
@@ -75,5 +95,14 @@ py_wheel(
     version = "{BUILD_EMBED_LABEL}",
     deps = [
         ":ncore_pkg",
+        "//tools/data_converter/pai:pai_converter_pkg",
```
andmax: Adding //tools/data_converter/pai:pai_converter_pkg to the wheel deps means the wheel installs a top-level tools/ package, the built wheel's top-level entries are ['ncore', 'nvidia_ncore-<ver>.dist-info', 'tools'].  tools is about as generic a name as exists on PyPI, so nvidia-ncore[pai] would collide with anything else installing a tools module in the same environment, and would shadow it depending on install order. Since this is the release that first publishes it, it's much cheaper to namespace now (e.g. under ncore/) than after users depend on the current layout.

ncore/BUILD.bazel
eattia-nvidia: Agreed that the extra has to match what the installed converter actually imports.

`scipy` and `universal_pathlib` are already required by the base `nvidia-ncore` wheel, so `pip install "nvidia-ncore[pai]"` gets them without listing them again on the extra.

`rich` is only used by `pai_remote/downloader.py` (and the `pai-clip-dl` CLI). Those are not part of `ncore-convert`: streaming conversion uses `config` / `index` / `remote` / `streaming`, and local conversion reads `pai-clip-dl` output from disk. `downloader.py` is no longer packaged, so the extra does not need `rich`. `pai-clip-dl` stays a Bazel-only tool in this repo.

ncore/BUILD.bazel
eattia-nvidia: Good catch — a top-level `tools` package on PyPI is too generic.

The published wheel now installs the converter under `ncore.converters` (`ncore.converters.pai.converter:cli`). There is no top-level `tools` in the wheel. In-repo Bazel targets are unchanged (`//tools/data_converter/pai:convert`); only the packaged import path is namespaced.

bazel/packaging/namespace_converter_srcs.py
```diff
@@ -0,0 +1,85 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
eattia-nvidia, line 1: What do you think about moving stuff into ncore/converters  to avoid namespace remapping ? @janickm

ncore/BUILD.bazel
andmax: True, thanks.

janickm:
I have some big reservations on this change and the whole repo and wheel structure is explicitly designed to _not_ ship (the wheel is for APIs, not for tools - the converters are merely "samples" of how to convert common data) - there might be different options here, let's discuss this offline first, I'll reach out (making draft for now)

janickm:
closing as discussed offline with the alternative bazel-based http_archive approach

PR #177, opened 2026-09-08
refactor(data): return Self from every CameraModelParameters.transform override

janickm:
`CameraModelParameters.transform` was declared four different ways across the hierarchy:

| class | declaration |
|---|---|
| `CameraModelParameters` (abstract) | `self: CameraModelParametersSelf` -> `CameraModelParametersSelf` (ad-hoc type-var) |
| `PinholeCameraModelParameters` | `self` -> `Self` |
| `FThetaCameraModelParameters` | `self` -> `FThetaCameraModelParameters` |
| `OpenCVFisheyeCameraModelParameters` | `self` -> `OpenCVFisheyeCameraModelParameters` |

The last two narrow `self` relative to the base, so those overrides are not substitutable for it. Two consequences: a helper generic over camera model parameters cannot call `transform` on them at all, and the return type collapses to the declaring class instead of preserving the caller's own type.

This unifies all four on PEP 673 `Self` and drops the now-unused `CameraModelParametersSelf` type-var. `Self` was already used in this module, and the module's `from __future__ import annotations` keeps it safe on the Python versions that do not provide it.

The regression test pins the property that motivates the change: a helper generic over the abstract base

```python
_T = TypeVar("_T", bound=CameraModelParameters)

def _downscale(parameters: _T, image_domain_scale: float) -> _T:
    return parameters.transform(image_domain_scale)
```

now type-checks and returns the caller's concrete type, verified for all four concrete parameter classes.

### Scope, explicitly

This does **not** make the same helper work when its type variable is bound to `ConcreteCameraModelParametersUnion` rather than the abstract base. I verified that separately: with a union bound the receiver is the whole union and is not assignable to any single member's `self`, so the call fails no matter how the overrides are annotated. `PinholeCameraModelParameters` demonstrates this independently of this change, since it already returned `Self` and still produced a diagnostic under a union bound.

Measured against a reproduction of that pattern, this change takes it from 4 diagnostics to 3: only the `invalid-return-type` one is fixed. The three remaining `invalid-argument-type` diagnostics are inherent to the union bound. Binding the type variable to `CameraModelParameters` instead resolves all of them, and does so on `main` today.

### DIFF
--- a/ncore/impl/data/types_test.py
+++ b/ncore/impl/data/types_test.py
@@ -16,14 +16,36 @@
 import io
 import unittest
 
-from typing import Optional
+from typing import List, Optional, Tuple, TypeVar, Union
 
 import numpy as np
 import numpy.testing as npt
 import PIL.Image as PILImage
 
 from ncore.impl.common.transformations import PoseGraphInterpolator
-from ncore.impl.data.types import EncodedImageData, PointCloud
+from ncore.impl.data.types import (
+    CameraModelParameters,
+    EncodedImageData,
+    FThetaCameraModelParameters,
+    IdealPinholeCameraModelParameters,
+    OpenCVFisheyeCameraModelParameters,
+    OpenCVPinholeCameraModelParameters,
+    PointCloud,
+    ShutterType,
+)
+
+
+#: A caller-side generic helper over camera model parameters. Bound to the abstract base, this is
+#: the shape downstream code needs in order to transform parameters while keeping the caller's
+#: concrete type; it only type-checks because every `transform` override returns `Self`.
+_CameraModelParametersT = TypeVar("_CameraModelParametersT", bound=CameraModelParameters)
+
+
+def _downscale(
+    parameters: _CameraModelParametersT,
+    image_domain_scale: Union[float, Tuple[float, float]],
+) -> _CameraModelParametersT:
+    return parameters.transform(image_domain_scale)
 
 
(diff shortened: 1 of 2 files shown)


ncore/impl/data/types_test.py
```diff
@@ -355,5 +377,65 @@ def test_memoization_survives_interleaved_instances(self) -> None:
         self.assertIs(a1, a2)
 
 
+class TestCameraModelParametersTransformSelf(unittest.TestCase):
+    """`transform` must preserve the concrete parameter type, statically and at runtime"""
+
+    @staticmethod
+    def _all_parameters():
```
janickm: should this have type annotations also? otherwise, why is this a method / it could be part of test_transform_preserves_concrete_type inlined?

janickm: Annotated, and kept as a helper since two tests use it.

The reason it was unannotated is worth recording: the honest annotation is `-> List[CameraModelParameters]`, and that breaks the second test, which assigns the result to `FThetaCameraModelParameters` to pin the `Self` behaviour. An element of an abstractly-typed list is not statically concrete.

So the fixtures are split: `_ftheta() -> FThetaCameraModelParameters` for the test that needs a concrete static type, and `_all_parameters() -> List[CameraModelParameters]` built on top of it for the runtime loop over all four. Both annotated, no duplicated construction.

PR #178, opened 2026-09-09
feat(sensors): open lidar model factory and make LidarModel generic in its parameters

janickm:
Applies the treatment the camera model hierarchy received in #175 / #177 to the lidar hierarchy. Same shape, so the review should feel familiar.

### The closed dispatch table, twice

`LidarModel.maybe_from_parameters` was a static method on the abstract base holding a closed `if`/`elif` table over every concrete subclass — and `StructuredLidarModel` **shadowed it with a second copy** that had to be kept in sync by hand. Since static methods do not dispatch, which of the two tables ran depended only on the class name the caller happened to spell, and neither could construct a lidar model defined outside NCore.

Both are replaced by a module-level `lidar_model_from_parameters()` dispatching on the parameter type via `functools.singledispatch`, with `maybe_lidar_model_from_parameters()` for the optional case and `register_lidar_model` for out-of-tree models. Both static methods are kept as deprecated forwarders.

### Generic model

`LidarModel` declared no `get_parameters` — only the concrete leaf did, so a `LidarModel`-typed value was a dead end. It is now generic in its parameter type and declares it. The parameter is **covariant**: it appears only in a return position, and invariance would leave concrete instantiations with no common `LidarModel[...]` supertype. Plain unparameterized `LidarModel` annotations and isinstance checks are unaffected.

### Serialization on the abstract base

`BaseLidarModelParameters` was not even an ABC (`@dataclass class BaseLidarModelParameters: pass`); the JSON mixin and `type()` sat only on the concrete leaf, so code holding the abstract type could not serialize. It now carries both, `type()` non-abstract so pre-existing out-of-tree subclasses stay instantiable, and the leaf drops its own mixin.

### One thing worth a look

The deprecated `StructuredLidarModel.maybe_from_parameters` keeps the **base's** parameter type rather than narrowing it to the structured parameters. Narrowing it is a contravariance violation that `ty` rejects once the base is widened — it was invisible before only because both copies happened to name the same concrete union. The structured-ness of the result is enforced at runtime instead.

Verified: 33/33 tests on 3.8 and 3.11, `ty` aspect clean, `//:format.check` clean.

### DIFF
--- a/ncore/impl/sensors/lidar_test.py
+++ b/ncore/impl/sensors/lidar_test.py
@@ -13,13 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import dataclasses
 import itertools
 import json
 import os
 import unittest
 import unittest.mock
 
-from typing import Tuple
+from typing import List, Tuple, Union, cast
 
 import numpy as np
 import parameterized
@@ -27,10 +28,17 @@ import torch
 
 from ncore.impl.common.transformations import se3_inverse
 from ncore.impl.common.util import unpack_optional
-from ncore.impl.data.types import RowOffsetStructuredSpinningLidarModelParameters
+from ncore.impl.data.types import BaseLidarModelParameters, RowOffsetStructuredSpinningLidarModelParameters
 from ncore.impl.sensors import lidar as lidar_module
 from ncore.impl.sensors.common import to_torch
-from ncore.impl.sensors.lidar import RowOffsetStructuredSpinningLidarModel
+from ncore.impl.sensors.lidar import (
+    LidarModel,
+    RowOffsetStructuredSpinningLidarModel,
+    StructuredLidarModel,
+    lidar_model_from_parameters,
+    maybe_lidar_model_from_parameters,
+    register_lidar_model,
+)
 
 
 # =============================================================================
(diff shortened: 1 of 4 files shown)


ncore/impl/sensors/lidar_test.py
```diff
@@ -73,6 +81,93 @@ def test_reserialization(self):
...
+
+        reference = self._parameters()
+        parameters = CustomLidarModelParameters(**dataclasses.asdict(reference))
+
+        # Without a registration, dispatch falls back to the base's factory along the MRO
+        self.assertIsInstance(
+            lidar_model_from_parameters(parameters, device="cpu"), RowOffsetStructuredSpinningLidarModel
+        )
+
+        @register_lidar_model
```
janickm, line 137: is it necessary to also have some @overload for the custom model or is this fine type wise?

janickm: Good question, and the answer is no for the test but yes as a documented caveat, because I checked what a type checker actually infers and it is not what I assumed.

`register_*` is purely a runtime mechanism; it cannot extend the overloads of the public entry point. So for a registered out-of-tree type:

- parameters deriving from the **abstract base** match the generic overload and resolve to `LidarModel[TheirParameters]`. Precise enough, no overload needed.
- parameters deriving from a **concrete** NCore parameters class (which is what this test does) match *that class's* overload. `reveal_type` gives `RowOffsetStructuredSpinningLidarModel`, not `CustomLidarModel`, even though the registered factory is what runs.

That stays sound here only because `CustomLidarModel` is a `RowOffsetStructuredSpinningLidarModel`. Register a model that is not, and the static type is wrong while the runtime is right. An out-of-tree user cannot fix that with an overload either, since they cannot add overloads to our function; they would need their own typed wrapper.

So: no overload in the test, but the caveat is now spelled out on `register_lidar_model`, and the test carries a comment recording the static-vs-runtime split so it does not read as an oversight. I mirrored the same wording onto `register_external_distortion_model` in #179 and, for consistency, onto the already-merged `register_camera_model` (that last one is a docstring-only change, carried in #179 since it already touches `camera.py`).

PR #180, opened 2026-09-09
feat(data): carry the concrete type inside serialized external distortion parameters

janickm:
Follow-up to #179, which introduced the `ExternalDistortionParameters` base but had to leave `CameraModelParameters.external_distortion_parameters` naming the concrete union. This removes that compromise.

### Why the field could not name the abstract base

`dataclasses_json` reconstructs whatever type a field is annotated with, and the serialized form carried no discriminator: the type was written *beside* the camera parameters by `encode_camera_model_parameters`, not inside the nested object. Confirmed directly:

```
>>> camera.to_dict()["external_distortion_parameters"].keys()
['horizontal_poly', 'horizontal_poly_inverse', 'reference_poly', 'vertical_poly', 'vertical_poly_inverse']
```

With a single-member union that resolves by luck. Against the abstract base, `from_dict` builds the base itself, which is a runtime fault rather than a typing one — it surfaced downstream as `TypeError: Unsupported external distortion type <class '...ExternalDistortionParameters'>`, only at the point of use.

### What changes

The serialized form becomes a discriminated union: the concrete `type()` is written inside the object under `EXTERNAL_DISTORTION_TYPE_KEY`, and the field's decoder dispatches on it through a registry that out-of-tree parameter types join via `register_external_distortion_parameters`, mirroring the model-side registry added in #179.

The field is then declared against `ExternalDistortionParameters`, and the four `get_parameters` casts name it too.

### Compatibility

Both directions hold, and each is covered by a test rather than assumed:

| case | behaviour |
|---|---|
| payload predating the nested key | decodes as `bivariate-windshield` — unambiguous, it was the only concrete type |
| previous encoded layout (type beside the camera parameters) | `decode_camera_model_parameters` migrates it into the nested object |
| reader predating this change, reading data written by it | still resolves the concrete type: `encode_` keeps writing the camera-level type |
| reader predating this change, seeing the added nested key | ignores it (verified against the published wheel) |

The camera-level type is therefore written twice for now. That is deliberate and commented; it can be dropped once no reader predating this change remains, and doing so would be the only breaking step — so it is not taken here.

### Scope

This widens a field that callers **read**, so downstream consumers passing it to concretely-typed APIs need updating in step with the release. The remaining read positions (`decode_*` return types and the parameter accessors in the compat layers) are deliberately left for a separate change, so this one stays a self-contained format change.

Verified: 33/33 tests on 3.8 and 3.11 with `--nocache_test_results`, `ty` aspect clean, `//:format.check` clean.

### DIFF
--- a/ncore/impl/data/types.py
+++ b/ncore/impl/data/types.py
@@ -21,10 +21,11 @@ import math
 import sys
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from enum import IntEnum, auto, unique
 from typing import (
     TYPE_CHECKING,
+    Any,
     Callable,
     ClassVar,
     Dict,
@@ -34,6 +35,8 @@ from typing import (
     Optional,
     Protocol,
     Tuple,
+    Type,
+    TypeVar,
     Union,
 )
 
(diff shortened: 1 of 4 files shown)


ncore/impl/data/types.py
```diff
@@ -875,15 +944,18 @@ def decode_camera_model_parameters(encoded_parameters: Mapping) -> ConcreteCamer
...
 
-    # Hook up typed external distortion type, if present
-    external_distortion_type: Optional[str] = encoded_parameters.get("external_distortion_type")
-    if external_distortion_type is not None:
-        if external_distortion_type == "bivariate-windshield":
-            camera_model_parameters["external_distortion_parameters"] = BivariateWindshieldModelParameters.from_dict(
-                camera_model_parameters["external_distortion_parameters"]
-            )
-        else:
+    # Older payloads carry the external distortion type beside the camera parameters rather than
```
janickm, line 947: add a unit test that validates both the old and new structure

janickm: Added `test_decode_accepts_old_and_new_payload_structures`: it builds both layouts from the same camera and asserts they decode to the *same* parameters, which is the property that actually matters here and which the existing tests did not cover -- they exercised each layout separately.

There were already five tests around this, all in `camera_test.py` next to the windshield fixtures rather than in `types_test.py`, which is probably why they were easy to miss:

- `test_serialized_form_carries_the_concrete_type` -- the tag is written inside the nested object
- `test_from_dict_reconstructs_the_concrete_distortion_type` -- new layout round-trips through the raw `from_dict` path
- `test_untagged_legacy_payload_still_decodes` -- nested object with no tag
- `test_legacy_camera_level_type_is_migrated_by_the_decoder` -- old layout, type beside the camera parameters
- `test_unknown_serialized_type_raises`

Also unquoted five stray `"ExternalDistortionParameters"` forward references in the same push. They were unnecessary on both counts: the module has `from __future__ import annotations`, and the class is defined above every use, so even the `TypeVar(bound=...)` runtime argument does not need the string.

PR #181, opened 2026-09-09
feat(skills): add the ncore-data-conversion agent skill

sanchitgarg:
## Description

Adds an agent skill for NCore V4 data conversion: running the six built-in
converters, adapting the nearest one for a dataset NCore does not support, and
checking the resulting store.

`SKILL.md` points at the in-tree documentation rather than restating it, and
carries only what `docs/` does not: half-open interval semantics, the FTheta
principal-point offset, `BBox3` centroid semantics, motion decompensation, the
projector's `--lidar-model` / `--external-distortion` defaults, and the
cross-converter sensor-id differences. Detail lives in `references/`.

Diagnosis is organised around separating the causes of misaligned
lidar-to-camera projections, since extrinsics, intrinsics and timestamps all
produce the same appearance and only the choice of frames tells them apart.

## Layout

```
skills/ncore-data-conversion/
├── SKILL.md
├── skill-card.md
├── evals/evals.json                   6 tasks (4 positive, 2 negative)
└── references/
    ├── v4-invariants.md
    └── adapting-a-converter.md
```

## Notes for reviewers

- **Licensed Apache-2.0**, matching the repository. The content is
  NVIDIA-authored throughout and vendors nothing third-party, so the earlier
  dual CC-BY-4.0 licensing and the per-directory `LICENSE`/`NOTICE` files were
  dropped. Keeping a second license at the repo root also breaks GitHub's
  license detection.
- **No CI workflow.** NCore should not push into the NVSkills pipeline; that
  side can pull. The dispatch also cannot work here as written: the
  `NVSKILLS_CI_DISPATCH_TOKEN` secret is not configured on this repo, `@main`
  on a third-party reusable workflow leaves ncore's CI surface under external
  control, and `on: pull_request` conflicts with the copy-pr-bot model that
  `ci.yml` relies on.
- **Named `ncore-data-conversion`**, not `ncore`. The bare slug collides with
  the sibling skill name already referenced from
  `physical-ai-neural-reconstruction` in the NVSkills catalog, and catalog
  entries must be unique.
- Frontmatter follows the agent-skills schema: `name`, `description`,
  `license`, `metadata`, with `version` nested under `metadata`. `tools:` was
  dropped as it is in no published schema.

Two bugs found while fact-checking the skill against the source are split out:
#182 (COLMAP `--start-time-sec` never reaches camera frames) and #183 (stale
converter test targets and defaults).

## Type of Change

- [x] Documentation update

## Checklist

- [x] I have read the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] My commits are GPG-signed
- [ ] I have added tests — n/a; `evals/evals.json` is the skill's evaluation set
- [x] New and existing tests pass locally (`bazel test //tools/...`)
- [x] Code is formatted (`bazel run //:format`)
- [x] I have updated documentation as needed
- [x] My changes include SPDX license headers on all new files

### DIFF
--- a/skills/ncore-data-conversion/SKILL.md
+++ b/skills/ncore-data-conversion/SKILL.md
@@ -1,0 +1,343 @@
+---
+name: ncore-data-conversion
+description: >-
+  Use when getting sensor data into an NCore V4 store: running a built-in
+  converter (PAI, Waymo, COLMAP/ScanNet++, KITTI, nuScenes, Argoverse 2),
+  adapting the nearest built-in converter for a dataset NCore does not ship
+  one for, or checking a converted store before something downstream reads
+  it. Covers V4 conventions for poses, camera and lidar models, cuboids and
+  timestamps. Do NOT use to train reconstructions or to extract per-object
+  3D assets.
+license: Apache-2.0
+metadata:
+  author: NVIDIA NCore
+  version: "0.3.0"
+  tags: ncore, data-conversion, sensors, v4, zarr, itar
+  upstream: https://github.com/NVIDIA/ncore
+---
+
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# NCore V4 data conversion
+
+Get sensor data into an **NCore V4** store, and check it before anything
+downstream consumes it. A V4 store is a general-purpose sensor-data format:
+NuRec is one consumer, so are `ncore_vis`, gsplat-based research code, and
+your own tooling. Nothing here assumes a NuRec pipeline.
+
+Three jobs, in order of preference: run a built-in converter; adapt the
+nearest built-in converter when none fits; check the result either way.
+
+This skill ships inside the ncore repository, so paths like
+`docs/data/conventions.rst` refer to files you already have. Prefer them over
+the rendered site, since they match your checkout.
+
+## Prerequisites
+
+- **Bazel**, via `bazelisk`. Not optional: converters and inspection tools are
+  Bazel targets. The `nvidia-ncore` wheel packages only the `ncore` library and
+  declares no console entry points, so `ncore_vis` and friends are reachable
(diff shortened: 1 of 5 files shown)


sanchitgarg:
@janickm The copy-PR security gate is holding the GPU CI; could you please vet this PR with `/ok-to-test` and review when convenient? Conventional Commits is green and the PR is mergeable.

janickm:
I took a pass at this and force-pushed onto your branch, based on offline discussion.

The content was well-researched; most of what I checked held up against the source. Main changes:

- **Fixed six factual errors.** The FTheta one mattered most: the skill said the runtime adds 0.5 to give `[959.5, 539.5]` for 1920x1080, but that is the *stored* value - runtime is `[960.0, 540.0]`. Half a pixel out, in exactly the direction the convention exists to prevent. Also `--world-global-mode` is waymo+colmap (only `localized` is Waymo-only) and the other four write `world_global` unconditionally; `:pytest_utils` is not a real target (the macro suffixes `_3_11`); doc URLs were missing the `data/` prefix; the `n_columns` vs `n_columns - 1` "inconsistency" is a closed vs half-open interval, each correct for its consumer.
- **Cut ~230 lines of duplication.** Pose conventions, camera models, spinning-lidar params, component layout and repo gates are all already in `docs/` and `CONTRIBUTING.md`. Since the skill ships in-tree it can just point at them, and they stay version-locked with the checkout. What is left is the material `docs/` genuinely does not carry.
- **Reworked the troubleshooting section.** It was a flat list keyed on "smeared" and "mirrored", which are not really descriptions of anything. Misaligned projections are the actual symptom, extrinsics/intrinsics/timestamps all produce it, and it is usually the source data rather than the conversion. Projecting at standstill separates calibration from timing, since timestamp errors have no geometric effect without ego motion. `--pose start` vs `end` is the second lever.
- **Apache-2.0 only** and dropped the CI workflow - see the description for the reasoning on both.

Two real bugs fell out of the fact-check: #182 and #183.

janickm (approved):

PR #184, opened 2026-09-11
feat(data): export the abstract lidar model parameter bases

janickm:
`ncore.data` exports `RowOffsetStructuredSpinningLidarModelParameters` and
`ConcreteLidarModelParametersUnion`, but none of the abstract bases above them. A consumer writing
a protocol or a helper over lidar parameters therefore has to name the single concrete class, even
when it only needs the shared surface.

The camera side has exported its abstract `CameraModelParameters` since 19.6.0, so this is an
asymmetry rather than a deliberate restriction.

All three levels are exported, not just the root, because each is a meaningful constraint:

| base | adds |
|---|---|
| `BaseLidarModelParameters` | the root: `type()` and the serialization surface |
| `BaseSpinningLidarModelParameters` | `spinning_frequency_hz`, `spinning_direction` |
| `BaseStructuredSpinningLidarModelParameters` | `n_rows`, `n_columns` |

A consumer that needs the row and column counts can require the level that declares them, rather
than the concrete class or the bare root.

## On `ConcreteLidarModelParametersUnion`

Worth recording, since it reads as a union but is not one:

```python
>>> import ncore.data as d
>>> d.ConcreteLidarModelParametersUnion is d.RowOffsetStructuredSpinningLidarModelParameters
True
```

It is declared `Union[RowOffsetStructuredSpinningLidarModelParameters]`, and Python collapses a
single-member union to the member itself, so it is that class at runtime and in the type checker.
It buys nothing over naming the concrete type, and it cannot grow a second member without
silently widening every annotation that uses it.

Left in place here: dropping a public export is a separate, breaking decision.

## Verification

- `bazel run //:format.check` clean; the ty lint aspect passes on all targets
- `__all__` checked for duplicates, and every listed name confirmed imported
- Behaviour confirmed against the published `nvidia-ncore==19.6.0` wheel in a clean venv

### DIFF
--- a/ncore/data/__init__.py
+++ b/ncore/data/__init__.py
@@ -46,6 +46,7 @@ from ncore.impl.data.types import (
     LabelSource,
     LabelType,
     LabelUnit,
+    LidarModelParameters,
     OpenCVFisheyeCameraModelParameters,
     OpenCVPinholeCameraModelParameters,
     PointCloud,
@@ -53,6 +54,8 @@ from ncore.impl.data.types import (
     ReferencePolynomial,
     RowOffsetStructuredSpinningLidarModelParameters,
     ShutterType,
+    SpinningLidarModelParameters,
+    StructuredSpinningLidarModelParameters,
     register_external_distortion_parameters,
 )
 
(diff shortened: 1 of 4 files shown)


ncore/data/__init__.py
```diff
@@ -26,6 +26,9 @@
     SequenceLoaderProtocol,
 )
 from ncore.impl.data.types import (
+    BaseLidarModelParameters,
```
janickm: should we rename `BaseLidarModelParameters` -> `LidarModelParameters` to be consistent with `CameraModelParameters` / `ExternalDistortionParameters`, which are also abstract / bases? I think this is ok as these are internal APIs currently and we should make this now when going public

PR #186, opened 2026-09-11
feat(data): open the model parameter hierarchies to out-of-tree implementations

janickm:
`IdealPinholeCameraModelParameters.from_source` and `natural_fov` dispatched over a hard-coded
`isinstance` chain of the four in-tree models. That makes 19.6.0's `register_camera_model` only half
an extension point: an out-of-tree model can be registered and built, then both helpers reject it,
with no way to opt in short of patching this module.

The per-model knowledge moves onto the models themselves. `CameraModelParameters.paraxial_pinhole_geometry`
returns the pinhole agreeing with the model to first order about the optical axis, and the helpers
consume it without knowing which model they were handed.

```python
@dataclass
class MyCameraModelParameters(CameraModelParameters):
    def paraxial_pinhole_geometry(self) -> ParaxialPinholeGeometry:
        return ParaxialPinholeGeometry(self.focal, self.centre, self.resolution)

# both now work, with nothing registered in this repository
IdealPinholeCameraModelParameters.from_source(my_params)
IdealPinholeCameraModelParameters.natural_fov(my_params)
```

## Why a named type rather than a tuple

`ParaxialPinholeGeometry` is the signature third parties implement. It should not depend on
remembering the order of three same-shaped arrays, and it can gain a field later without breaking
every implementer.

## Three implementations cover four models

`PinholeCameraModelParameters` already carries the focal length and principal point shared by the
ideal and OpenCV pinholes, so one override serves both. The OpenCV fisheye declares those fields
itself instead of inheriting them, so it repeats the body. F-Theta derives its focal from the
first-order term of the forward polynomial, as before.

## Requirements on concrete models

`paraxial_pinhole_geometry` and `type` are `@abstractmethod` on the parameter bases, so a new model
has to decide both rather than inheriting a silent refusal. A camera model with no paraxial pinhole
opts out by implementing the method and raising `TypeError`. `type` is abstract on all three bases,
not only the camera one, so the requirement is the same wherever a model is defined.

Concrete subclasses that implement neither raise `TypeError` on instantiation. Every model in this
repository does, and there is no out-of-tree model yet, which is the point: this is the moment to
state the contract, before one exists to break. Deliberately **not** marked as a breaking change,
since nothing outside this repository can be affected and it should not force a major version.

This replaces the defensive `raise NotImplementedError` defaults that `type` previously carried on
each base, whose docstrings justified themselves by protecting out-of-tree subclasses that do not
exist.

No public name is changed or removed; `ParaxialPinholeGeometry`, the registration functions and the
methods are additions.

## Tests

- `test_out_of_tree_model_opts_in_by_implementing_the_geometry` defines a model this repository
  knows nothing about and shows both helpers working on it, which is the whole point of the change.
- `test_paraxial_geometry_matches_the_in_tree_models` pins the reported geometry against what
  `from_source` then builds, for all four in-tree models.
- `test_out_of_tree_model_may_opt_out_by_raising` covers the opposite case: a model that declines,
  whose `TypeError` travels out through both helpers unchanged.
- `test_model_omitting_the_geometry_cannot_be_instantiated` pins that omitting it is not a way to
  decline, which is what makes the requirement real rather than advisory.

## Reading the parameters back

The factory change alone left registration half-finished. A model registered with
`register_camera_model` could be constructed and converted to a pinhole, but never read back from
storage, because `decode_camera_model_parameters` dispatched over a hard-coded `if/elif` of the four
in-tree identifiers (and the lidar decoder over its single one).

Both decoders now resolve through a registry, the same shape the external distortion parameters have
used since their type key was introduced. `register_camera_model_parameters` and
`register_lidar_model_parameters` are the deserialization counterparts to the factories'
`register_camera_model` / `register_lidar_model`.

**No serialized format change.** `encode_*` already wrote the identifier from `type()`, so existing
data reads exactly as before. Verified across versions rather than argued: a payload encoded by the
published `nvidia-ncore==19.6.0` wheel decodes under this branch to the same class and the same
values, both as written and with the identifier rewritten to the retired `pinhole`.

The `pinhole` identifier is the one thing not recoverable from `type()`: it was the name of what is
now `opencv-pinhole`. It moves into an explicit alias table rather than vanishing into the registry
keys, so the reason it exists stays visible.

## Coverage: legacy, current, out-of-tree

| case | test |
|---|---|
| current identifiers, all in-tree camera models | `test_current_camera_identifiers_round_trip` |
| **legacy** `pinhole` payload | `test_legacy_pinhole_identifier_still_decodes` |
| unknown identifier still raises, naming it | `test_unknown_camera_identifier_raises_naming_it` |
| **out-of-tree** model encodes and decodes | `test_out_of_tree_camera_parameters_round_trip` |
| **out-of-tree** model usable by the pinhole helpers | `test_out_of_tree_model_opts_in_by_implementing_the_geometry` |
| out-of-tree model opting out by raising | `test_out_of_tree_model_may_opt_out_by_raising` |
| model omitting it entirely is not instantiable | `test_model_omitting_the_geometry_cannot_be_instantiated` |
| each base requires its own `type` | assertions in the three model-identifier tests |
| duplicate identifier rejected, in-tree class keeps it | `test_duplicate_camera_registration_is_rejected` |
| re-registering the same class is a no-op | `test_registering_the_same_class_twice_is_idempotent` |
| current lidar identifier round-trips | `test_current_lidar_identifier_round_trips` |
| **out-of-tree lidar** parameters round-trip | `test_out_of_tree_lidar_parameters_round_trip` |
| **out-of-tree external distortion** round-trips inside a camera | `test_out_of_tree_external_distortion_round_trips_inside_a_camera` |
| unknown lidar identifier raises, naming it | `test_unknown_lidar_identifier_raises_naming_it` |
| in-tree geometry matches what `from_source` builds | `test_paraxial_geometry_matches_the_in_tree_models` |
| geometry rejects wrong dtype / shape / non-positive focal | `TestParaxialPinholeGeometry` |

The legacy `pinhole` alias is the regression most likely to slip through a move to a registry, since
it is the one identifier the old chain knew that `type()` does not report, hence its own test.

All three hierarchies now have an out-of-tree round trip, not just the camera one. The external
distortion path is the odd one: its parameters are not encoded standalone but as a type-tagged
object nested inside the camera parameters, so the tag has to survive `encode_camera_model_parameters`
and resolve on the way back. Each was confirmed non-vacuous by removing the registration decorator
and watching it fail with `Unknown ... type`.

## No private helper left

`_paraxial_geometry` had nothing to do once models declared their own geometry, so it is gone and
both consumers ask the source directly. Its `isinstance` guard went with it, so an argument that is
not `CameraModelParameters` at all now fails as an `AttributeError` rather than a `TypeError`. That
case only ever arose from a deliberate `cast`, and the test asserting it is removed rather than
adjusted: it pinned the interpreter's behaviour for an argument the signature forbids, not a promise
of this module.

## One drive-by

`BivariateWindshieldModelParameters` carried `@dataclass` outside `@register_...`, so registration
saw the class before `@dataclass` had processed it. That is harmless today, since `@dataclass`
returns the same object, but it would silently register a stale class if anyone ever added
`slots=True` or `frozen=True`. The new registrations put the decorator outside, and this one is
reordered to match.

## Verification

- ty lint aspect clean on every target; `bazel run //:format.check` clean

Squashed to a single commit on `main`.


ncore/impl/data/types.py
janickm: why not using the new type here - this is a private method anyway, we can change it's signature.

actually, can't we get rid of this altogether as this is already implemented as `paraxial_pinhole_geometry` / refactor the "consumers" to use `paraxial_pinhole_geometry()`?

ncore/impl/data/types.py
```diff
@@ -243,6 +244,21 @@ def __post_init__(self) -> None:
...
+
+    The paraxial approximation a model is willing to make of itself: the pinhole agreeing with it
+    to first order about the principal direction. It is what
+    :meth:`IdealPinholeCameraModelParameters.from_source` builds on, and what
+    :meth:`IdealPinholeCameraModelParameters.natural_fov` measures.
+    """
+
+    focal_length: np.ndarray  #: Per-axis focal length ``[fu, fv]`` [px] (float32, [2,])
+    principal_point: np.ndarray  #: Principal point in image coordinates [px] (float32, [2,])
+    resolution: np.ndarray  #: Image resolution ``[width, height]`` [px] (uint64, [2,])
```
janickm, line 256: should we have a _post_init checking types and shapes?

ncore/impl/data/types.py
janickm: Gone entirely in c348ee7d, along with the tuple. Both consumers now call
`source.paraxial_pinhole_geometry()` directly; the helper was only wrapping that one call and
unpacking it.

One consequence worth calling out, because it is a behaviour change rather than a pure refactor.
The helper also carried the `isinstance(source, CameraModelParameters)` guard, so an argument that
is not camera parameters at all now fails as an `AttributeError` instead of a `TypeError`.

I removed `test_unsupported_source_raises` rather than adjusting it. It passed
`cast(ConcreteCameraModelParametersUnion, object())`, so it pinned the interpreter's behaviour for
an argument the signature already forbids, not anything this module promises. The case that does
matter, a real `CameraModelParameters` subclass that never overrides the method, still raises a
`TypeError` from the base default and is covered by `test_out_of_tree_camera_parameters_raise`.

Say the word if you would rather keep the guard, and I will put a single `isinstance` back at the
top of `from_source` instead.

ncore/impl/data/types.py
janickm: Added in c348ee7d, in the same shape the concrete parameter classes use: dtype and shape per field,
plus focal positivity.

It earns its place more here than on an in-tree class, since these values come from an out-of-tree
implementation. Without it a float64 focal or a stray third element surfaces much later, inside the
focal-length arithmetic in `from_source`, where the traceback points at our code rather than at the
implementation that produced the bad value.

Covered by `TestParaxialPinholeGeometry`: wrong dtype, wrong shape and non-positive focal, per field.

ncore/impl/data/types.py
```diff
@@ -261,6 +291,22 @@ class CameraModelParameters(dataclasses_json.DataClassJsonMixin, ABC):
...
+        :func:`ncore.sensors.register_camera_model` becomes usable by both by implementing this, and
+        needs nothing else.
+
+        The default refuses rather than guessing, so a model with no meaningful paraxial pinhole
+        fails with a clear error instead of an attribute lookup on a field it does not have.
+
+        Raises:
+            TypeError: If the concrete model does not define a paraxial pinhole.
+        """
+        raise TypeError(f"Unsupported camera model type for ideal pinhole conversion: {type(self).__name__}")
```
janickm: let's make this fully abstract also / require downstream implementations where they can opt out if needed (same as `transform()` - let's require this by default)

PR #188, opened 2026-09-15
feat(data): declare the field-of-view contract on the lidar parameter base

janickm:
Every lidar has a vertical and a horizontal field of view, whatever its scanning design, the same
way every camera has a resolution. Until now both were declared only on
`RowOffsetStructuredSpinningLidarModelParameters`, so there was no way to ask an arbitrary lidar for
its field of view, and a second model could arrive with a different spelling or none at all.

`LidarModelParameters` now declares both.

## The implementation stays where it is

This is the part that differs from the camera equivalent, where the geometry helper could be lifted
to a shared base. The lidar implementation derives the vertical extent from `row_elevations_rad` and
the horizontal one from `column_azimuths_rad` and `row_azimuth_offsets_rad`, and all three exist only
on the row-offset model, not on `StructuredSpinningLidarModelParameters` above it.

A flash or solid-state lidar would derive the same two quantities from something else entirely. That
is precisely the argument for putting the contract on the base and leaving the derivation to each
model.

## On the `dtype` parameter

Kept. It reads oddly on something declared as a base-level property, and it was worth reconsidering
while changing the signature's home. But it is the released spelling, and callers pass a precision
matched to the model they are building, so narrowing it would break them for no benefit.

## Requirement on concrete models

Concrete lidar parameter classes must implement both, as they already must implement `type`. Every
model in this repository does, and there is still no out-of-tree lidar model to accommodate, which is
the moment to state a contract rather than discover later that it is missing.

Deliberately **not** marked as a breaking change: nothing outside this repository can be affected, so
it should not force a major version.

## Tests

- `test_out_of_tree_model_supplies_its_own_field_of_view` defines a flash lidar with no rows, columns
  or rotation, and shows it reporting a field of view. That is the case the change exists for, and it
  is not expressible against the old arrangement.
- `test_model_omitting_the_field_of_view_cannot_be_instantiated` asserts the failure names both
  missing methods, which is what makes the requirement real rather than advisory.
- The existing out-of-tree lidar registry fixture now implements the contract too, which is a small
  demonstration that the requirement reaches every model that registers.

## Verification

- ty lint aspect clean on every target; `bazel run //:format.check` clean


ncore/impl/data/types_test.py
```diff
@@ -577,6 +578,12 @@ class _OutOfTreeLidarModelParameters(LidarModelParameters):
             def type() -> str:
                 return "out-of-tree-lidar-test"
 
+            def get_vertical_fov(self, dtype: Any = np.float32) -> util.FOV:
```
janickm, line 581: the type is wrong and not used?

PR #190, opened 2026-09-18
feat(sensors): add the ideal orthographic camera model

janickm:
## What

Adds an **ideal orthographic camera model** (`ideal-orthographic`): a
distortion-free parallel projection that drops the camera-frame depth instead of
dividing by it, so a point's image location does not depend on it.

This is NCore's first **non-central** camera model. Its rays are parallel and
share no common origin, so a 3d direction no longer identifies a ray:
unprojection returns 6d `[origin, direction]` rays, matching the layout
`WorldRaysReturn` already uses. `CameraModel.camera_ray_dim` reports which
representation a model uses (`3` or `6`).

## Why

Bird's-eye-view raster data is naturally described by a parallel projection, and
NCore had no way to express one: FTheta, pinhole and fisheye are all central
projections. Without it, consumers hand-roll their own projection alongside
NCore's camera models, which then does not participate in the shared pose,
rolling-shutter and serialization machinery.

Orthographic is a standard projection rather than a bespoke one, e.g. OpenUSD's
`GfCamera::Projection {Perspective, Orthographic}` and OpenGL's `glOrtho`.

## Design notes

**The ray interface is asymmetric, which keeps the change small.**
`camera_rays_to_image_points` receives an unnormalized camera-frame *point* for
every model (for a central model a point and a direction coincide up to scale;
a non-central model projects the point itself). So the `world_points_to_*`
family and the rolling-shutter solver work unchanged, and **rolling shutter
comes for free**. Only unprojection needed generalizing.

**Central models are provably unaffected.** Per-ray origins are now carried
through the sensor pose as `R @ origin + t` instead of broadcasting the pose
translation. A central model's origins are zero, so the same expression
reproduces the previous broadcast exactly, with no branch. There is a regression
test asserting this for all three central models.

**Two combinations are refused**, because a parallel bundle makes them
ill-defined rather than merely unimplemented:

- *External distortion* deflects each ray individually and would not preserve
  the parallel bundle. Rejected at construction: a ray-dimensionality guard
  alone would let the pairing survive every forward projection and only surface
  on unprojection.
- *Rectification against a central model* has no depth-independent solution, as
  there is no common centre to pivot about. Rectifying two non-central models is
  well-defined but not implemented yet (`NotImplementedError`).

`paraxial_pinhole_geometry()` raises via the documented opt-out: an orthographic
camera's focal length is infinite, so it has no pinhole approximation and
`IdealPinholeCameraModelParameters.from_source()` correctly refuses it.

**Naming.** `pixels_per_unit` is the orthographic counterpart of a pinhole's
`focal_length`, occupying the same place in the projection but differing in
dimension: a pinhole consumes the dimensionless `x/z`, this consumes `x`
directly. Named per-*unit* rather than per-*meter* because coordinates may be
`UNITLESS` (e.g. SfM reconstructions). The `Ideal` prefix mirrors
`IdealPinholeCameraModelParameters` and leaves the unqualified `Orthographic`
family name free for a distorted telecentric sibling, avoiding a repeat of the
retired `pinhole` -> `opencv-pinhole` identifier alias.

**No serialized format changed.** The model registers through the existing
`register_camera_model` / `register_camera_model_parameters` hooks, and the
discriminator was already written from `type()`.

## Also in this PR

- **Renamed the forward projection** to `camera_points_to_image_points` /
  `camera_points_to_pixels` (old names kept as documented-deprecated
  forwarders). See below.
- `ExternalDistortionModel` now guards the ray representation in its public
  methods and dispatches to `_impl` hooks, mirroring `CameraModel`. See the note
  below.
- `camera_ray_dim` is a `ClassVar`, since the ray representation is a property of
  the model type rather than of an instance.
- Extracted the image-domain scale/resolution prelude that was duplicated across
  all four `transform()` implementations (and dropped a redundant double
  `.astype(np.uint64)` in the FTheta one).

## Renaming `camera_rays_to_image_points`

Raised in review. The forward method's argument was always an unnormalized
camera-frame **point**, never a ray. With only central models the distinction
was unobservable, because every central projection is scale-invariant:

| Model | Forward math | Magnitude |
|---|---|---|
| ideal / OpenCV pinhole | `cam_rays[:,:2] / cam_rays[:,2:3]` | cancels |
| FTheta / OpenCV fisheye | `atan2(xy_norm, z)`, then `delta/xy_norm * xy` | cancels |
| **ideal orthographic** | `cam_rays[:,:2] * pixels_per_unit` | **is the answer** |

A non-central projection has no such freedom: the point's `x` and `y` are
precisely what is being projected, so passing a normalized direction silently
yields a different image location rather than an error. That promotes the name
from merely loose to actively misleading, so it is fixed here rather than after
release:

- `camera_rays_to_image_points` -> `camera_points_to_image_points`
- `camera_rays_to_pixels` -> `camera_points_to_pixels`

Old names remain as forwarders carrying a `.. deprecated::` directive, matching
the docstring-only deprecation style already used in this module (there is no
runtime `warnings.warn` anywhere in the repo, so none was introduced).

Note the *unprojection* side keeps its name: `image_points_to_camera_rays`
genuinely returns rays, and for a non-central model those rays are the whole
point of the 6d representation.

## Note on the `ExternalDistortionModel` refactor

The refactor renames the two abstract methods an external distortion model
implements (`distort_camera_rays` / `undistort_camera_rays` ->
`_distort_camera_rays_impl` / `_undistort_camera_rays_impl`), so that the public
methods can hold the ray-representation guard.

Not flagged as a breaking change: there are no out-of-tree implementations of
the abstract base, and models deriving from a concrete class such as
`BivariateWindshieldModel` inherit the new hooks unchanged. Were an out-of-tree
subclass of the abstract base to exist, it would fail loudly at instantiation
(`Can't instantiate abstract class ...`) rather than silently.

## Testing

- Projection: window-fraction correctness across the full extent, depth
  invariance, out-of-window validity, and that points *behind* the camera stay
  valid (no frustum, unlike every other model).
- 6d rays: distinct origins with a shared direction, and the
  unprojection/projection round trip.
- World rays: per-ray origins under a rotated and translated pose, for both the
  static and rolling-shutter paths; plus a regression guard that the three
  central models are bit-identical.
- Guards: distortion rejected at construction, `paraxial_pinhole_geometry()` and
  `from_source()` raising, non-central rays rejected by the distortion model,
  and both `Rectificator` cases.
- Serialization round trip, `transform()`, and factory dispatch.
- Point semantics: central projections are scale-invariant, the orthographic one
  is not, and the deprecated aliases agree with the renamed methods. The alias
  test covers the orthographic model specifically, since it is the only one
  whose result would change if an alias perturbed its argument in passing.

`bazel test //... --config=no-gpu` passes (36/36, Python 3.8 and 3.11),
`bazel build //...` is clean under the `ty` aspect, and `bazel run
//:format.check` reports no violations.

## Docs

New "Central and Non-Central Models" and "Ideal Orthographic Camera Model"
sections in `sensor_models.rst`, a "Camera Rays" section in `conventions.rst`
documenting the 3d/6d representations, caveats in the external-distortion and
rectification sections, and `formats.rst` now lists `ideal-orthographic` (plus
the previously missing `ideal-pinhole`).


docs/data/conventions.rst
```diff
@@ -139,6 +139,31 @@ start with ``[0.0, 0.0]`` at the top-left corner of the top-left pixel in the
...
+
+Unprojecting an image point yields a *camera ray* in the extrinsic camera frame.
+Its representation depends on whether the camera model is central:
+
+* **Central models** (FTheta, ideal pinhole, OpenCV pinhole, OpenCV fisheye)
+  project through a single centre. A ray is a 3d direction ``[x, y, z]``; its
+  origin is implied to be the camera frame's origin.
+* **Non-central models** (the ideal orthographic model) have rays that share no
+  common origin. A ray is 6d, an origin followed by a direction,
+  ``[ox, oy, oz, dx, dy, dz]``.
```
janickm: these are still relative to the camera frame origin, or? we should document this

docs/data/conventions.rst
```diff
@@ -139,6 +139,31 @@ start with ``[0.0, 0.0]`` at the top-left corner of the top-left pixel in the
...
+
+:attr:`~ncore.sensors.CameraModel.camera_ray_dim` reports which a model uses.
+World rays (as returned by the
+``image_points_to_world_rays_*`` and ``pixels_to_world_rays_*`` methods) are
+always 6d ``[origin, direction]``, for every model.
+
+The *forward* direction is uniform: ``camera_rays_to_image_points`` takes a 3d
+camera-frame **point** regardless of the model, because a central model's point
```
janickm: it's point or direction (normalized point), or (both relative to the camera origin)?

janickm: or does it strictly need to be a point for non-central models? in that case the method name is a little misleading, but maybe we should accept this for now?

janickm: actually, maybe the name of the method was bad all along and we should rename (maybe with a deprecation of the current one?)

ncore/impl/sensors/camera.py
```diff
@@ -325,6 +366,53 @@ def __init__(
...
+        # and only surface on unprojection.
+        if self.external_distortion is not None and self.camera_ray_dim != _CENTRAL_CAMERA_RAY_DIM:
+            raise TypeError(
+                f"{type(self).__name__} is a non-central camera model and cannot carry external "
+                "distortion: deflecting each ray individually would not preserve the parallel ray "
+                "bundle the model is defined by."
+            )
+
+    @property
+    def camera_ray_dim(self) -> int:
```
janickm: should this be a class-wide or even static property?

ncore/impl/sensors/camera.py
```diff
@@ -325,6 +366,53 @@ def __init__(
...
+        :class:`CameraModel.WorldRaysReturn`.
+
+        Note this describes :meth:`image_points_to_camera_rays` (unprojection) only. The forward
+        direction, :meth:`camera_rays_to_image_points`, takes a 3d camera-frame *point* for every
+        model: for a central model a point and a direction coincide up to scale, and a non-central
+        model projects the point itself.
+        """
+        return _CENTRAL_CAMERA_RAY_DIM
+
+    def _split_camera_rays(self, camera_rays: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
janickm, line 400: see the other note on the current bad name of camera_ray vs point (as these are really points, or?)

docs/data/conventions.rst
janickm: Yes, and you're right that it wasn't stated. Both representations are in the extrinsic camera frame: a central model's 3d direction is measured from the camera frame origin, and a non-central model's explicit origin is a point in that same frame. Rewritten to open with that, so the frame is established before either representation is described.

docs/data/conventions.rst
janickm: Point. I checked each model and the docs were misleading in both directions:

| Model | Forward math | Magnitude |
|---|---|---|
| ideal / OpenCV pinhole | `cam_rays[:,:2] / cam_rays[:,2:3]` | cancels |
| FTheta / OpenCV fisheye | `atan2(xy_norm, z)`, then `delta/xy_norm * xy` | cancels |
| ideal orthographic | `cam_rays[:,:2] * pixels_per_unit` | **is the answer** |

So every central projection is scale-invariant and a direction works fine, while the orthographic one requires an actual point. Docs now state the asymmetry explicitly, with a warning that normalizing the input for a non-central model silently moves the result. Added tests pinning both halves, which nothing covered before.

ncore/impl/sensors/camera.py
janickm: Class-wide, agreed — it never depends on instance state. Now a `ClassVar[int]`, following the precedent at `types.py:1568`. Verified it works on an `nn.Module` subclass: readable from both class and instance, doesn't land in `state_dict()`, survives `.to()`. Test added.

ncore/impl/sensors/camera.py
janickm: The rename above covers the forward direction. I left `_split_camera_rays` as-is: it operates on the output of `image_points_to_camera_rays`, which really are rays, so its name is accurate. Happy to rename if you'd still prefer.

docs/data/conventions.rst
janickm: Yes, strictly. The orthographic projection is `x, y -> image`, so normalizing rescales exactly the quantity being projected. It fails silently, not loudly, which is what tipped me from "loose name" to "needs renaming".

janickm: Agreed, and done: `camera_rays_to_image_points` -> `camera_points_to_image_points` and `camera_rays_to_pixels` -> `camera_points_to_pixels`, with the old names kept as forwarders carrying `.. deprecated::`. Docstring-only, matching the existing style in this module — the repo has no runtime `warnings.warn` anywhere, so I didn't introduce one.

One asymmetry worth noting: only the *forward* name was wrong. `image_points_to_camera_rays` genuinely returns rays, and for a non-central model those rays are the whole point of the 6d representation, so it keeps its name.

